PR #16, opened 2022-04-24
feature-representation_metric

PososikTeam:


src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+                    query_idxs = np.arange(i, i + search_batch_size)
+                    
+
+                closest_dist, closest_idx = index.search(queries[query_idxs], k=k)
+                relevant = relevants[query_idxs]
+ 
+                if self.metric == 0:
+                    closest_dist = 1 - closest_dist
+            
+                closest = map(lambda idx: np.stack((db_ids[closest_idx[idx]], closest_dist[idx]), axis=1), \
```
VladVin, line 303: Do we really need map here? May be `db_ids[closest_idx]` will still be correct and faster

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+            vectors: Vectors of all storage, include queries and database vectors, size (database_size, embedding_size).
+            queries_idxs: Bool array wich indicate if vector is query, size (database_size).
+            scores: Array of scores for every relevant index per each query request, size (database_size, query_size).
+                See representation dataset for more information.
+
+        Retruns:
+            q_vecs: Queries vectors, size (queries_size, embedding_size).
+            db_vecs: Database vectors, for faiss build index, size (database_size, embedding_size).
+            relevant: Array of relevant indexes in database for every query request, size (queries_size, ).
+            scores: Array of scores without queries empty scores.
```
VladVin, line 162: "scores: Array of scores related to queries that have at least one relevant item."

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+                closest_dist, closest_idx = index.search(queries[query_idxs], k=k)
+                relevant = relevants[query_idxs]
+ 
+                if self.metric == 0:
+                    closest_dist = 1 - closest_dist
+            
+                closest = map(lambda idx: np.stack((db_ids[closest_idx[idx]], closest_dist[idx]), axis=1), \
+                    np.arange(len(closest_idx)))
+                
+                if scores is None:
```
VladVin, line 306: Should be mentioned in docstring that `scores` is optional

VladVin (changes requested):

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+
+        Output in relevant array, contain it's index in database and score for current query.
+        Output in closest array, contain it's index in database and distance for current query.
+        
+        Args:
+            index: Faiss database built index.
+            relevants: Relevant indexes for every query, size (query_size, ) and the second shape is can be different
+                for every query request.
+            queries: Vectors for every query request, size (query_size, embedding_size).
+            scores: Scores for every relevant index per each query request, size (database_size, query_size).
```
VladVin, line 250: `size (database_size, total_num_queries)`

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+                if self.metric == 0:
+                    closest_dist = 1 - closest_dist
+            
+                closest = map(lambda idx: np.stack((db_ids[closest_idx[idx]], closest_dist[idx]), axis=1), \
+                    np.arange(len(closest_idx)))
+                
+                if scores is None:
+                    relevant = map(lambda r: np.stack((r, np.ones_like(r)), axis=1), relevant)
+                else:
+                    relevant = map(lambda r_q: np.stack((r_q[0], scores[r_q[0], q_order_idxs[r_q[1]]]), axis=1), \
```
VladVin, line 309: Shouldn't scores be filtered like this? 
```
query_scores = scores[r_q[0], q_order_idxs[r_q[1]]]
rel_scores = query_scores[query_scores > 0.]
```

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+            self.add_state('queries_idxs', default=[], dist_reduce_fx=None)
+            self.add_state('scores', default=[], dist_reduce_fx=None)
+
+    def update(self, vectors: torch.Tensor, targets: Optional[torch.Tensor] = None, \
+               queries_idxs: Optional[torch.Tensor] = None, scores: Optional[torch.Tensor] = None):
+        """Append tensors in storage.
+        
+        Args:
+            vectors: Often it would be embeddings, size (batch_size, embedding_size).
+            targets: The vectors targets, size (batch_size).
```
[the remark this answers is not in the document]
PososikTeam: It's size for tensor, which equal (batch_size)

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+                        zip(relevant, query_idxs))
+                
+                relevant = list(relevant)
+                closest = list(closest)
+                yield relevant, closest
+
+        return generator()
+
+    def build_index(self, vectors: np.ndarray):
+        """Build index with faiss of a given set of vectors.
```
VladVin, line 319: Build index of a given set of vectors with FAISS.

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+
+        Args:
+            vectors: Database vectors on with is built faiss index, size (database_size, embedding_size).
```
VladVin, line 322:
```suggestion
            vectors: Database vectors to index, size (database_size, embedding_size).
```

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+        queries_idxs = np.delete(queries_idxs, empty_relevant_idxs)
+
+        return q_vecs, db_vecs, relevant, scores, db_idxs, queries_idxs
+
+    def prepare_classification_data(self, vectors: np.ndarray, targets: np.ndarray
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """Prepare data for faiss build index, and following search, in classification dataset case.
+        
+        Separate query and database vectors from storage vectors.
+        Query vectors index would be first found target uniq value in targets array. Query size is uniq target count.
```
VladVin, line 194: "Query vector index will be the first row index for each unique target value found in `targets` array. So, total number of queries is equal to unique target count. The remaining vectors will be treated as database vectors (again for each unique target value)"

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+
+        Args:
+            vectors: Vectors of all storage, include queries and database vectors, size (database_size, embedding_size).
+            targets: Targets in classification task for every vectors, size (database_size).
```
VladVin, line 198:
```suggestion
            targets: Targets in classification task for every vector, size (database_size)
```
Also, add range for values

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+            q_vecs, db_vecs, relevants, scores, \
+                db_idxs, q_order_idxs = self.prepare_representation_data(vectors, queries_idxs, scores)
+
+        q_vecs = q_vecs.astype(np.float32)
+        db_vecs = db_vecs.astype(np.float32)
+
+        index = self.build_index(db_vecs)
+
+        # if search batch size is None, search queries vectors by one request
+        search_batch_size = len(q_vecs) if self.search_batch_size is None else self.search_batch_size
```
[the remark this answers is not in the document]
PososikTeam: It might be overload the memory, if request size is large and vectors sizes is also large. What do you suggest? Set as default search_batch_size = num (8)?

VladVin (changes requested):

src/metrics/representation.py
PososikTeam: It's already done in function `self.prepare_representation_data`.

src/metrics/representation.py
PososikTeam: If we not use map, we need create list then stack db_ids with closest_distance.

src/metrics/representation.py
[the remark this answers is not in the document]
VladVin: Yes, something like this. As far as I know, `compute` is called in a single process. So, you can safely set `batch_size` equal to number of CPUs

VladVin (changes requested):

tests/metrics/test_representation_ddp.py
```diff
@@ -0,0 +1,230 @@
...
+            TestCase(test_name='average_precision', dataset='representation'),
+            TestCase(test_name='ndcg', dataset='representation'),
+        ]
+
+        for case in test_cases:
+            actual = run_model(case)
+            name = case.class_name
+            for k in range(1, 5):
+                assert math.isclose(case.expected[k], actual[k])
+            # print(f'Answer = {actual}')
```
VladVin, line 219: Remove

src/metrics/representation.py
```diff
@@ -0,0 +1,388 @@
...
+        return index
+
+
+# @METRICS.register_class
+class PrecisionAtKMeter(IndexBasedMeter):
+    def compute(self):
+        scores = []
+        generator = super().compute()
+        for relevant_idx, closest_idx in generator:
+            scores += precision(relevant_idx, closest_idx, k=self.k).tolist()
```
VladVin, line 356: Since only the function is changed, you can create a callable `metric_func` and change it in the inherited classes

VladVin (changes requested):

src/metrics/representation.py
```diff
@@ -0,0 +1,381 @@
...
+        k: Number of top closest indexes to get.
+        search_batch_size: The size for one faiss search request.
+        normalize_input: If true vectors woulds be normalize, overwise no.
+
+    Raises:
+        ValueError: If metric or dataset is not correct write.
+    """
+    metrics = {'IP': 0, 'L2': 1}
+    dataset = {'classification': 0, 'representation': 1}
+    def __init__(self, exact_index: bool, dataset: Union[str, int], metric: Union[str, int], k: int = None, \
```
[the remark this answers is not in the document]
VladVin: 4. Not solved

tests/metrics/test_representation.py
```diff
@@ -0,0 +1,214 @@
...
+
+import torch
+from torchmetrics import Metric
+from typing import List, Dict, Union
+import numpy as np
+from src.metrics.representation import RecallAtKMeter, PrecisionAtKMeter, MeanAveragePrecisionAtKMeter, NDCGAtKMeter
+from src.metrics.representation import DatasetType, MetricDistance
+
+
+vectors = {
```
VladVin, line 11: You can create `utils.py` and move every repeated code blocks there (DRY principle)

VladVin (changes requested):

PR #19, opened 2022-04-27
create feature hooks base model

PososikTeam:


VladVin (changes requested):

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+        input_forward_reductions: Optional value of input reduction list for forward. Need to reconstruct initial input
+            tensor shape.
+        input_hooks_channels: Optional value of input channels list for hooks.
+        input_hooks_reductions: Optional value of input hooks reduction list. Need to reconstruct initial input tensor 
+            shape.
+    """
+    def __init__(self, input_forward_channels: List[int], input_forward_reductions: List[int] = None, \
+                 input_hooks_channels: List[int] = None, input_hooks_reductions: List[int] = None):
+        super().__init__()
+        self._feature_info = None
```
VladVin, line 120: Add a public method `get_features_info` and return None by default, so that a user will have to implement it in order to access feature hooks. Call it in the `create_hooks` method. Describe it clearly in the docstring.

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+        self._input_forward_reductions = input_forward_reductions
+        self._input_hooks_channels = input_hooks_channels
+        self._input_hooks_reductions = input_hooks_reductions
+
+        self._output_forward_channels = None
+        self._output_forward_reductions = None
+        self._output_hooks_channels = None
+        self._output_hooks_reductions = None
+
+    def create_hooks(self):
```
VladVin, line 132: 1. This method has to be called on a model initialization stage. So, it needs to be clearly explained in the docstring. Otherwise, user will not know about this functionality.
2. Move arguments from `__init__` to `create_hooks` because it's related to the same functionality which isn't used by default. Also wrap these parameters into a list of `FeaureInfo`, it will be easier for user to pass

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        for feature in self._feature_info:
+            if type(feature) != FeatureInfo:
+                raise ValueError('All feature_info must be FeatureInfo class.')
+
+    def forward(self, x: torch.Tensor):
+        # Return features for classification.
+        raise NotImplementedError
+
+    def forward_backbone_features(self, x: torch.Tensor) -> Tuple[torch.Tensor, List[torch.Tensor]]:
```
VladVin, line 156: Since the model is just universal `BaseModel` there shouldn't be word `backbone`. So I suggest you to join methods `forward_backbone_features` and `forward_stage_features` into a single one and call it `forward_stage_features`

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+        backbone_features = list(backbone_features.values())
+        backbone_features = [x] + backbone_features
+        return last_features, backbone_features
+
+    def forward_stage_features(self, x: torch.Tensor) -> List[torch.Tensor]:
+        """Return only hooks features."""
+        x = self.forward(x)
+        return list(self._feature_hooks.get_output(x.device).values())
+
+    def init_weights(self):
```
VladVin, line 169: Why is it here?

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+        """Generate feature hooks."""
+        # Check if all self._feature_info is FeatureInfo types.
+        self.check_feature_info_type(self)
+        self._stage_names = [feature.module_name for feature in self._feature_info]
+        self._output_hooks_channels = [feature.channel_number for feature in self._feature_info]
+        self._output_hooks_reductions = [feature.reduction for feature in self._feature_info]
+        hooks = [Hook(module_name=name, hook_type=HookType.FORWARD) for name in self._stage_names]
+        self._feature_hooks = FeatureHooks(hooks, self.named_modules())
+
+    def check_feature_info_type(self):
```
VladVin, line 142: This should be a private method

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+        self._output_hooks_reductions = None
+
+    def create_hooks(self):
+        """Generate feature hooks."""
+        # Check if all self._feature_info is FeatureInfo types.
+        self.check_feature_info_type(self)
+        self._stage_names = [feature.module_name for feature in self._feature_info]
+        self._output_hooks_channels = [feature.channel_number for feature in self._feature_info]
+        self._output_hooks_reductions = [feature.reduction for feature in self._feature_info]
+        hooks = [Hook(module_name=name, hook_type=HookType.FORWARD) for name in self._stage_names]
```
VladVin, line 139: Hook type is hardcoded, but the method `create_hooks` is general for any model

VladVin (changes requested):

src/models/base_model.py
```diff
@@ -0,0 +1,226 @@
...
+
+        self._output_forward_channels = None
+        self._output_forward_reductions = None
+        self._output_hooks_channels = None
+        self._output_hooks_reductions = None
+
+    def create_hooks(self):
+        """Generate feature hooks."""
+        # Check if all self._feature_info is FeatureInfo types.
+        self.check_feature_info_type(self)
```
VladislavPatrushev, line 135: Send two 'self'

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        It must be rewrite in inherited classes, if Module would need hooks.
+
+        Retruns: Hooks FeatureInfo list.
+        """
+        return None
+
+    def _create_hooks(self) -> Tuple[List[FeatureInfo], FeatureHooks]:
+        """Call self.get_features_info() method to generate feature_info list and then generate feature hooks.
+        
+        If self.get_features_info() not rewrited in inheritable class the features_info and hooks would be None.
```
VladVin, line 137: If `self.get_features_info()` isn't overrided in an inherited class then the `features_info` and `hooks` will be None and no feature hooks will be created.

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+            ValueError: If rewrited self.get_features_info() method return not FeatureInfo list.
+        """
+        super().__init__()
+
+        # Create feature_info list with hooks
+        self._features_info, self._feature_hooks = self._create_hooks()
+
+
+    def get_features_info(self) -> List[FeatureInfo]:
+        """Method for initialize Hooks.
```
VladVin, line 126: Initialize feature hooks

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        
+        It must be rewrite in inherited classes, if Module would need hooks.
```
VladVin, line 128: It should be overriden in an inherited class if a user is expected to have access to feature hooks of the implemented Module

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+
+        return features_info, feature_hooks
+
+    def __check_features_info_types(self):
+        """Check if all self._feature_info is FeatureInfo class.
+            
+        Raises:
+            ValueError: If any value in feature_info is not FeatureInfo class.
+        """
+        for feature in self._features_info:
```
VladVin, line 163: `self._features_info` doesn't exist at this point. Pass as argument

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+            if type(feature) != FeatureInfo:
+                raise ValueError('All feature_info must be FeatureInfo class.')
+
+    def forward(self, x: torch.Tensor):
```
VladVin, line 167: I don't think you need to restrict `forward` method to accept only `x` argument. Just remove this method and a user will implement `Module`'s interfaces which is broader than this

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        # Return features for classification.
+        raise NotImplementedError
+
+    def forward_stage_features(self, x: torch.Tensor) -> Tuple[torch.Tensor, List[torch.Tensor]]:
```
VladVin, line 171: Make the interface like `Module.forward` method

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        """Return model output with hooks features.
+        
+        Returns:
+            last_features: Module forward method output.
+            hooks_features: Hooks outputs.
+        """
+        last_features = self.forward(x)
+        hooks_features = self._feature_hooks.get_features(x.device)
+        hooks_features = list(hooks_features.values())
+        hooks_features = [x] + hooks_features
```
VladVin, line 181: I don't think you should add `x` to list of outputs. Since the user passed `x` as input he already has this value

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        return last_features, hooks_features
+
+    def _get_output_hooks_channels(self) -> List[int]:
```
VladVin, line 184: `_get_hooks_output_channels`

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @abstractmethod
+    def _get_output_forward_channels(self) -> Union[int, List[int]]:
+        """Set output channels for Module forward pass.
+        
+        Returns: Outpus channels.
+        """
+        pass
+    
+    def get_output_channels(self) -> Tuple[Union[int, List[int]], List[int]]:
```
VladVin, line 201: Why is it public? Do you expect a user to override it?

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        Returns:
+            forward_channels: Forward pass output channels.
+            hooks_channels: Hooks output channels.
+        """
+        forward_channels = self._get_output_forward_channels()
+        hooks_channels = self._get_output_hooks_channels()
+        return forward_channels, hooks_channels
+
+    @property
+    def feature_info(self):
```
VladVin, line 213: `features_info`

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        """Generate hooks output channels numbers.
+        
+        Returns:
+            output_hooks_channels: Hooks output channels numbers.
+        """
+        output_hooks_channels = [feature.num_channels for feature in self._feature_info]
+        return output_hooks_channels
+
+    @abstractmethod
+    def _get_output_forward_channels(self) -> Union[int, List[int]]:
```
VladVin, line 194: `get_forward_output_channels`

VladVin (changes requested):

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        self._hook_features[device] = OrderedDict()  
+        return output
+
+
+class BaseModel(nn.Module, ABC):
+    """Base model for all TorchOk Models - Backbone, Neck, Pooling and Head.
+
+    This class supports adding feature hooks to some model layers.
+    Class has feature info list for every hook.
+    To create hooks, method self.get_features_info() must be rewrited!
```
VladVin, line 110: "must be overriden"

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        # clear after reading
+        self._hook_features[device] = OrderedDict()  
+        return output
+
+
+class BaseModel(nn.Module, ABC):
+    """Base model for all TorchOk Models - Backbone, Neck, Pooling and Head.
+
+    This class supports adding feature hooks to some model layers.
+    Class has feature info list for every hook.
```
VladVin, line 109: "Class has feature info to describe hooks."

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+    To create hooks, method self.get_features_info() must be rewrited!
+    Also class contains method get_output_channels wich returns forward pass and hooks output channels.
```
VladVin, line 111: "which"

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+    This class supports adding feature hooks to some model layers.
+    Class has feature info list for every hook.
+    To create hooks, method self.get_features_info() must be rewrited!
+    Also class contains method get_output_channels wich returns forward pass and hooks output channels. 
+    """
+    def __init__(self):
+        """Inits BaseModel class and it's hooks.
+
+        Raises:
+            ValueError: If rewrited self.get_features_info() method return not FeatureInfo list.
```
VladVin, line 117: "If overrided ..."

src/models/base_model.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        super().__init__()
+
+        # Create feature_info list with hooks
+        self._features_info, self._feature_hooks = self._create_hooks()
+
```
VladVin, line 123: Remove empty line

VladVin (changes requested):

src/models/base_model.py
PososikTeam: Because it will use in Task.

src/models/base_model.py
PososikTeam: I think it must be protected `_get_forward_output_channels`.

src/models/base_model.py
```diff
@@ -0,0 +1,245 @@
...
+            hook_name = hook.module_name
+            module = modules[hook_name]
+            hook_fn = partial(self._collect_hook_features, hook_name)
+            # register hooks
+            if hook.hook_type == HookType.FORWARD_PRE:
+                module.register_forward_pre_hook(hook_fn)
+            else:
+                module.register_forward_hook(hook_fn)
+           
+        self._feature_outputs = defaultdict(OrderedDict)
```
VladVin, line 72: Define `self.__feature_hooks` and `self.__features_info` because every field of the class should be defined in `__init__`

src/models/base_model.py
```diff
@@ -0,0 +1,245 @@
...
+        # Check if all self._features_info is FeatureInfo types.
+        if self._features_info is not None:
+            self.__check_features_info_types()
+            hooks = [
+                Hook(module_name=feature.module_name, hook_type=feature.hook_type) 
+                for feature in self._features_info
+                ]
+            self._feature_hooks = FeatureHooks(hooks, self.named_modules())
+
+    def __check_features_info_types(self):
```
VladVin, line 182: Make it static and pass `features_info` as argument. Reason: if things might go bad with types then `self._features_info` shouldn't be set beforehand

src/models/base_model.py
```diff
@@ -0,0 +1,245 @@
...
+                raise ValueError('All features_info must be FeatureInfo class.')
+
+    def forward_stage_features(self, *input: Any) -> Tuple[torch.Tensor, List[torch.Tensor]]:
+        """Return model output with hooks features.
+        
+        Returns:
+            last_features: Module forward method output.
+            hooks_features: Hooks outputs.
+        """
+        device = list(*input)[0].device
```
VladVin, line 199: Why not to use `self.device`?

src/models/base_model.py
VladVin: But a user has to implement this method according to that you put `pass` as an implementation

src/models/base_model.py
PososikTeam: nn.Module has no attribute device.

src/models/base_model.py
```diff
@@ -0,0 +1,234 @@
...
+        """Initialize feature hooks.
+        
+        It should be overriden in an inherited class if a user is expected to have access to feature hooks 
+        of the implemented Module.
+
+        Retruns: Hooks FeatureInfo list.
+        """
+        return []
+
+    def create_hooks(self, *args: Any, **kwargs: Any) -> Tuple[List[FeatureInfo], FeatureHooks]:
```
VladVin, line 149: Why does a user need to pass arguments here? I haven't seen any case requiring this

src/models/base_model.py
```diff
@@ -0,0 +1,245 @@
...
+    Example:
+        >>> class ModelWithHooks(BaseModel):
+        >>>     def __init__(self, input_channel: int):
+        >>>         super().__init__()
+        >>>         out_features = [5, 10, 15]
+        >>>         self.conv1 = nn.Conv2d(input_channel, 5, 3)
+        >>>         self.conv2 = nn.Conv2d(5, 10, 3)
+        >>>         self.conv3 = nn.Conv2d(10, 15, 3)
+        >>>         # Call _create_hooks after all modules is initialize, with parameters as in _get_features_info.
+        >>>         self._create_hooks(output_channels=[10, 15], module_names=['conv2', 'conv3'])
```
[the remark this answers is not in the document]
VladVin: Shouldn't not be any arguments here

src/models/base_model.py
VladVin: I don't get the point. While you are writing to the variable `self.*` you actually store the values in your class's field

tests/models/test_base_module.py
```diff
@@ -0,0 +1,134 @@
...
+
+import torch
+import torch.nn as nn
+
+from typing import List
+
+from src.models.base_model import BaseModel, FeatureInfo
+
+
+# class TestBaseModelWithoutHooks(unittest.TestCase):
```
VladVin, line 11: Restore

VladVin (changes requested):

src/models/base_model.py
PososikTeam: The reason is user's reluctance to put hook parameters into class fields

PR #25, opened 2022-05-04
Config structure

PososikTeam:


src/constructor/constructor.py
```diff
@@ -90,7 +91,14 @@ def __create_optimizer(parameters: Union[Module, Tensor, List[Union[Module, Tens
     def __create_scheduler(optimizer: Optimizer, scheduler_params: DictConfig) -> Dict[str, Any]:
         scheduler_class = SCHEDULERS.get(scheduler_params.name)
         scheduler = scheduler_class(optimizer, **scheduler_params.params)
-        pl_params = scheduler_params.pl_params if 'pl_params' in scheduler_params else {}
+        pl_params = scheduler_params.pl_params
+        # Omega conf error when create pl_params with only one argument, it's request second,
```
VladVin, line 95: Didn't get this comment, what is the argument?

src/constructor/config_structure.py
```diff
@@ -0,0 +1,125 @@
...
+class LossParams:
+    name: str
+    tag: str
+    mapping: Dict[str, str]
+    params: Dict = field(default_factory=dict)
+    weight: Optional[float] = None
+
+@dataclass
+class JointLossParams:
+    loss_params: List[LossParams]
```
VladVin, line 82: Let's rename it to `losses` and the outer name in config to `joint_loss`. It might be easier for a user to check if he needs to understand behaviour of the joint loss thing

src/metrics/metrics_manager.py
```diff
@@ -94,23 +45,21 @@ def compute(self):
...
+        """Initialize MetricManager.
+
+        Args:
+            params: Metric parameters.
+        """
         super().__init__()
         # Change list to set in phases
         for param in params:
-            param.phases = set(param.phases)
+            param.phases = list(set(param.phases))
```
VladVin, line 58: Do you still need set here?

tests/constructor/configs/config.yaml
```diff
@@ -0,0 +1,76 @@
+# defaults:
```
VladVin, line 1: What does this comment mean?

tests/constructor/test_config_structure.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import *
+
+from hydra.core.config_store import ConfigStore
+from omegaconf import OmegaConf
+
+from src.constructor.config_structure import ConfigParams, Phase
+
+
+@hydra.main(config_path="configs/", config_name="config.yaml")
+def test_config_load_when_full_config_was_define(config):
```
VladVin, line 16: was_defined

tests/constructor/test_config_structure.py
```diff
@@ -0,0 +1,56 @@
...
+    config_params = ConfigParams(**config)
+    # test if metric phase is enum
+    if isinstance(type(config_params.metrics[0].phases[0]), Phase):
+        print('Failed. Phase is not Enum.')
```
VladVin, line 20: This isn't a fail, this is just a warning. If you want the test to fail here, you should use an assert statement

tests/constructor/test_config_structure.py
```diff
@@ -0,0 +1,56 @@
...
+    else:
+        print('Phase enum OK.')
```
VladVin, line 22: I don't see any benefits of using prints in tests

tests/constructor/test_config_structure.py
VladVin: Also, the name of the test suggests a big test, but in fact only two parameters are checked. Shouldn't you check all the parameters that you expect with a bunch of assert statements? (and check the comment below)

tests/constructor/test_config_structure.py
```diff
@@ -0,0 +1,56 @@
...
+    # test normalize_weights parameter
+    if config_params.losses.normalize_weights != False:
+        print('Failed. normalize_weights != False.')
+    else:
+        print('normalize_weights OK.')
+
+
+@hydra.main(config_path="configs/", config_name="config_without_metrics.yaml")
+def test_config_load_when_metric_is_not_define(config):
```
VladVin, line 31: Maybe you should just leave one giant test as `test_config_load_when_full_config_was_defined` or you can split this one by many small tests where every group of parameters is checked independently, so that it will be easier for contributors to find errors

tests/constructor/test_config_structure.py
```diff
@@ -0,0 +1,56 @@
...
+    if len(params) == 0:
+        print('Load empty params is OK.')
+    else:
+        print('Failed. Load empty params.')
+
+
+if __name__ == "__main__":
+    cs = ConfigStore.instance()
+    cs.store(name="config", node=ConfigParams)
+    test_config_load_when_full_config_was_define()
```
VladVin, line 53: Looks like these tests won't be captured by unittest

tests/constructor/test_constructor.py
```diff
@@ -229,29 +232,32 @@ def __test_transforms_augmentations(self, kind):
...
-                    'params': {},
-                    'tag': 'classification',
-                    'mapping': {
-                        'input': 'logits',
-                        'target': 'target'
-                    },
-                    'weight': 0.7
-                }
-            ]
+            'losses':{
```
VladVin, line 235: Did you run `flake8 <path/to/source/file>` ?

tests/constructor/test_constructor_with_conf_structure.py
```diff
@@ -0,0 +1,228 @@
+import sys
```
VladVin, line 1: I see you pass the same structure with which `Constructor` was already tested in the `tests/constructor/test_constructor.py`. I think we don't need these tests, because config structure and constructor should be tested independently

train.py
```diff
@@ -0,0 +1,52 @@
...
+from hydra.core.config_store import ConfigStore
+from pathlib import Path
+
+
+# Hack to fix multiprocessing deadlock when PyTorch's DataLoader is used
+# (more info: https://github.com/pytorch/pytorch/issues/1355)
+import cv2
+cv2.setNumThreads(0)
+
+# from src.constructor import TASKS
```
VladVin, line 15: Remove

train.py
```diff
@@ -0,0 +1,52 @@
...
+from src.constructor.config_structure import ConfigParams
+# from src.constructor import create_trainer
+
+
+@hydra.main()
+def load_config(config):
+    # Need to add --config-path (-cp) and --config_name (-cn) in run command
```
VladVin, line 22: `config-name`

train.py
```diff
@@ -0,0 +1,52 @@
...
+    # Example config in configs/classification_cifar10.yaml
```
VladVin, line 23: Examples should be in `examples` folder

train.py
```diff
@@ -0,0 +1,52 @@
...
+    # Then the command line will be:
+    # python --config-path configs --config_name classification_cifar10 
+    config_params = ConfigParams(**config)
+    return config_params
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-cp', '--config-path', type=str,
```
VladVin, line 31: I think we should support just `config` parameter directing to the full config. It will be the only required argument

train.py
```diff
@@ -0,0 +1,52 @@
...
+    parser.add_argument('-cn', '--config-name', type=str,
+                        help="Yaml file name with configuration parameters.")
+    parser.add_argument('-jl', '--job-link', type=str,
+                        help="sagemaker job name, if running localy set to 'local'", default='local')
+    args = parser.parse_args()
+    
+    # getconfig name
+    config_name = args.config_name
+    config_name = config_name.split('.')[0] if '.' in config_name else config_name
+    structure_name = 'base_' + config_name
```
VladVin, line 42: Why do we need this?

train.py
```diff
@@ -0,0 +1,52 @@
...
+@hydra.main()
+def load_config(config):
+    # Need to add --config-path (-cp) and --config_name (-cn) in run command
+    # Example config in configs/classification_cifar10.yaml
+    # Then the command line will be:
+    # python --config-path configs --config_name classification_cifar10 
+    config_params = ConfigParams(**config)
+    return config_params
+
+if __name__ == '__main__':
```
VladVin, line 29: Can the config parameters be overriden by cli arguments?

train.py
VladVin, line 23: Put "TODO: add config into the project"

VladVin (changes requested):

tests/constructor/configs/config.yaml
PososikTeam: Same reason as ```'base_' + config_name```

https://hydra.cc/docs/upgrades/1.0_to_1.1/automatic_schema_matching/#:~:text=base_%20prefix%2C%20e.g.%20base_mysql.

train.py
PososikTeam: Yes hydra support change any config parameters by use cli

train.py
PososikTeam: Hydra not support only name it need path to

train.py
PososikTeam: Actually no =). I need to remove argparser to give hydra ability change config by cli

VladVin: Ok, let's do it then

train.py
VladVin: But you aren't using `hydra.main`. Ok, let's move to the correct usage of hydra

tests/constructor/test_hydra_load.py
```diff
@@ -0,0 +1,24 @@
...
+import omegaconf
+from omegaconf import OmegaConf, DictConfig
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import *
+
+from src.constructor.config_structure import ConfigParams, Phase
+
+
+@hydra.main(config_path='/workdir/rbayazitov/torchok/examples/configs/', config_name='example')
```
VladVin, line 14: Hard coded path.  
Overall, do you think we need this kind of test? Looks like you are testing Hydra's behaviour but not ours

tests/constructor/test_config_structure_load.py
```diff
@@ -0,0 +1,44 @@
...
+
+def load_structured_config(path):
+    config = OmegaConf.load(path)
+    OmegaConf.resolve(config)
+    schema = OmegaConf.structured(ConfigParams)
+    config = OmegaConf.merge(schema, config)
+    return config
+
+
+class TestConfigSrucure(unittest.TestCase):
```
VladVin, line 21: `TestConfigStructure`

train.py
```diff
@@ -0,0 +1,39 @@
...
+def main(config: DictConfig):
+    # Need to add --config-path (-cp) and --config_name (-cn) in run command
+    # Example config in configs/classification_cifar10.yaml
+    # TODO: add config into the project
+    # Then the command line will be:
+    # python --config-path configs --config_name classification_cifar10
+
+    # Resolve -> change evn variable to values for example ${oc.env:USER} -> 'root'
+    OmegaConf.resolve(config)
+    # Registrate structure
```
VladVin, line 26: Register

train.py
```diff
@@ -0,0 +1,39 @@
...
+    schema = OmegaConf.structured(ConfigParams)
+    # Merge sturcture with config
+    config = OmegaConf.merge(schema, config)
+    # may be we don't need this line
```
VladVin, line 30: Need or don't need?

VladVin (changes requested):

PR #30, opened 2022-05-18
Hot FIX Metric test

PososikTeam:


src/metrics/representation.py
```diff
@@ -195,18 +195,22 @@ def prepare_representation_data(self, vectors: np.ndarray, queries_idxs: np.ndar
         
         relevant = []
         empty_relevant_idxs = []
+        # print(f'scores = {scores}')
```
VladVin, line 198: Remove

src/metrics/representation.py
```diff
@@ -195,18 +195,22 @@ def prepare_representation_data(self, vectors: np.ndarray, queries_idxs: np.ndar
...
         for idx in range(len(q_vecs)):
             relevant_idxs = np.where(scores[:, queries_idxs[idx]] > 0.)[0]
             if len(relevant_idxs) == 0:
                 empty_relevant_idxs.append(idx)
             else:
+                # Need sort relevant indexes by its scores for NDCG metric
```
VladVin, line 204: Need to sort relevant indexes by its scores for NDCG metric

src/metrics/representation.py
```diff
@@ -195,18 +195,22 @@ def prepare_representation_data(self, vectors: np.ndarray, queries_idxs: np.ndar
...
+                current_scores = scores[relevant_idxs, queries_idxs[idx]]
+                sort_indexes = np.argsort(current_scores)
+                relevant_idxs = relevant_idxs[sort_indexes[::-1]]
```
VladVin, line 207: You sorted relevant indices but not scores

src/metrics/representation.py
```diff
@@ -195,18 +195,22 @@ def prepare_representation_data(self, vectors: np.ndarray, queries_idxs: np.ndar
...
                 relevant.append(relevant_idxs)
-        relevant = np.array(relevant)
-        
+        relevant = np.array(relevant, dtype=np.object)
```
VladVin, line 209: Not a good type. Why is it like that?

src/metrics/representation.py
```diff
@@ -320,8 +324,8 @@ def generator():
  
                 if self.metric_distance == MetricDistance.IP:
                     closest_dist = 1 - closest_dist
-            
-                closest = map(lambda idx: np.stack((db_ids[closest_idx[idx]], closest_dist[idx]), axis=1), \
+
+                closest = map(lambda idx: np.stack((db_ids[closest_idx[idx]], [1] * len(closest_idx[idx])), axis=1), \
```
VladVin, line 328: Why distances is always 1?

src/optim/optimizers/__init__.py
```diff
@@ -1,20 +1,20 @@
-from torch.optim import (
-    Adadelta, Adagrad, Adam, Adamax, AdamW, ASGD, LBFGS,
-    RMSprop, Rprop, SGD, SparseAdam
-)
+# from torch.optim import (
```
VladVin, line 1: What happens?

tests/metrics/representation/envaronment/envaronment.py
```diff
@@ -0,0 +1,103 @@
+import torch
```
VladVin, line 1: 1. envaronment -> environment
2. It's not an invironment, but rather context
3. You can leave these files in the same directory `tests/metrics/representation/`

VladVin (changes requested):

src/metrics/representation.py
PososikTeam: it's dosen't important, for NDCG score=distance need for sort more relevant examples, but in this part of code we had already sorted our examples by faiss. So if we change 1 o distance with type float the index of relevant will be also float and after that inside ranx it may be fail to compare relevant int index with our relevant float index

src/metrics/representation.py
PososikTeam: Doesn't need to sort the scores, because scores is the marix with query-relevant information, but it's good to sort relevant indexes beforehand, so that for every query we have sorted by score its relevant indexes.

src/metrics/representation.py
PososikTeam: It is some warning numpy just try it fix, but unsuccessfully, just forgot delete.

src/metrics/representation.py
```diff
@@ -320,8 +326,8 @@ def generator():
...
```
VladVin, line 330:
```suggestion
# NDCG score=distance is needed to sort more relevant examples, but in this part of code we had already sorted our
# examples by faiss. So if we change 1 to distance with type float the index of relevant will be also float and after that
# inside ranx it may be fail to compare relevant int index with our relevant float index
closest = map(lambda idx: np.stack((db_ids[closest_idx[idx]], [1] * len(closest_idx[idx])), axis=1), \
    np.arange(len(closest_idx)))
```

PR #32, opened 2022-05-20
Feature dataset cifar10

VladislavPatrushev:


src/data/datasets/cifar10.py
```diff
@@ -0,0 +1,128 @@
...
+    test_list = [
+        ['test_batch', '40351d587109b95175f43aff81a1287e'],
+    ]
+    meta = {
+        'filename': 'batches.meta',
+        'key': 'label_names',
+        'md5': '5ff9c542aee3614f3951f8cda6e48888',
+    }
+
+    def __init__(self,
```
VladVin, line 39: Put documentation

VladVin (changes requested):

PR #35, opened 2022-05-30
Feature Segmentation Dataset

VladislavPatrushev:


src/data/datasets/segmentation/image_segmentation.py
```diff
@@ -0,0 +1,117 @@
...
+            sample['mask'] = mask
+
+        sample = self._apply_transform(self.augment, sample)
+        sample = self._apply_transform(self.transform, sample)
+
+        sample['image'] = sample['image'].type(torch.__dict__[self._image_dtype])
+        sample['index'] = idx
+
+        if not self._test_mode:
+            sample['target'] = sample['mask'].type(torch.__dict__[self.__target_dtype])
```
VladVin, line 84: What if we leave the name `mask` for targets?

tests/data/datasets/test_image_segmentation.py
```diff
@@ -0,0 +1,59 @@
...
+class TestClassificationMulticlass(unittest.TestCase):
+    def __init__(self, methodName: str = None) -> None:
+        super().__init__(methodName)
+        self.__root_dir = Path(__file__).parent
+        self.__data_folder = self.__root_dir / 'data/segmentation_data/'
+        self.__csv_path = self.__root_dir / 'data/segmentation_data/segmentation_test.csv'
+        self.__transform = Compose([Resize(224, 224), ToTensorV2()], p=1.0)
+        self.__augment = Compose([HorizontalFlip(p=0.5), VerticalFlip(p=0.5)])
+
+    def test_len(self):
```
VladVin, line 21: The tests are partially the same as for other datasets. Create a general unit test for inheritance

VladVin (changes requested):

PR #36, opened 2022-05-31
SOP run

PososikTeam:


src/constructor/constructor.py
```diff
@@ -183,7 +183,9 @@ def __prepare_dataloader(dataset_params: DictConfig, dataloader_params: DictConf
     @staticmethod
     def __create_dataset(dataset_params: DictConfig) -> ImageDataset:
         transform = Constructor.__create_transforms(dataset_params.transform)
-        augment = Constructor.__create_transforms(dataset_params.augment)
+        # TODO remove when OmegaConf is fix bug, write to issue to Omegaconf!
```
VladVin, line 186: What's a bug? Explain in the comment. Always.

src/metrics/representation.py
```diff
@@ -13,15 +13,28 @@
 from ranx.metrics import precision, recall, average_precision, ndcg
 from enum import Enum
 
+from src.constructor import METRICS
+
 
 class DatasetType(Enum):
-    CLASSIFICATION = 0
-    REPRESENTATION = 1
+    CLASSIFICATION = 'cls'
```
VladVin, line 20: Not consistent. Either all the parameters are short names or all are full names

src/metrics/representation.py
```diff
@@ -13,15 +13,28 @@
...
+    REPRESENTATION = 'representation'
 
 
 class MetricDistance(Enum):
-    IP = 0
-    L2 = 1
+    IP = 'ip'
```
VladVin, line 25: Better to use big letters, otherwise might be not understood in another context

VladVin (changes requested):

src/tasks/classification.py
```diff
@@ -19,14 +18,19 @@ def __init__(self, hparams: DictConfig):
...
+
+        pooling_params = self._hparams.task.params.get('pooling_params', dict())
+        pooling_in_features = self.backbone.get_forward_output_channels()
+        pooling_name = self._hparams.task.params.get('pooling_name', 'IdentetyPooling')
+        self.pooling = POOLINGS.get(pooling_name)(in_features=pooling_in_features, **pooling_params)
+        
+        head_params = self._hparams.task.params.get('head_params', dict())
+        head_in_features = self.pooling.get_forward_output_channels()
+        # TODO write IdentetyHead
+        head_name = self._hparams.task.params.get('head_name', 'IdentetyHead')
```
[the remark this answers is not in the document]
PososikTeam: Because it's a self._hparams.task.params: Dict, it's not hard structure. May be we need change task config, and write config for backbone pooling and head.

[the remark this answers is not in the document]
VladVin: Ok, but why do you need TODO here?  
Also, `IdentityHead` is a correct name

PR #38, opened 2022-06-05
CIFAR-10 run

VladVin:


src/models/heads/base.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, in_features, out_features):
         self._out_features = out_features
 
     @abstractmethod
-    def forward(self, *args, **kwargs) -> torch.Tensor:
+    def forward(self, *args, **kwargs) -> Tensor:
```
VladislavPatrushev, line 23: We need to decide on a codestyle for tensor. I wrote:
def forward(self, *args, **kwargs) -> torch.Tensor:

PR #39, opened 2022-06-06
Load checkpoint

PososikTeam:


src/tasks/base.py
```diff
@@ -101,12 +102,15 @@ def __check_drop_last_params(self, data_params: List[Dict[str, Any]], phase: str
                 raise ValueError(f'DataLoader parametrs `drop_last` must be False in {phase} phase.')
 
     def on_train_start(self) -> None:
-        # TODO check and load checkpoint
-        pass
+        if self.current_epoch == 0:
+            load_checkpoint(self, base_ckpt_path=self.hparams.task.base_checkpoint, 
+                            override_name2ckpt_path=self.hparams.task.override_checkpoints,
```
VladislavPatrushev, line 107: hparams is protected self._hparams.task.base_checkpoint e.t.c.. It's not good call property in class

tests/constructor/test_load_checkpoint.py
```diff
@@ -0,0 +1,251 @@
...
+
+    model_keys = [
+        'layer1.module.conv1.weight', 
+        'layer1.linear.weight',
+        'linear.weight',
+    ]
+    
+
+    def test_generate_require_state_dict_when_only_base_checkpoint_was_defined(self):
+        base_state_dict = {
```
VladislavPatrushev, line 56: DRY. In several test function i saw base_state_dict with identity keys and values. Move to class property( like model_keys)

VladVin (changes requested):

src/constructor/load.py
```diff
@@ -0,0 +1,220 @@
...
+                raise ValueError(f'Error in load checkpoint function. '
+                                 f'Key `{straighten_key}` in ovrride checkpoint not in model state dict')
+            weight = state_dict[override_key]
+            state_dict_with_depth.add(straighten_key, weight, depth)
+
+    override_state_dict = state_dict_with_depth.state_dict
+    return override_state_dict
+
+
+def generate_require_state_dict(model_keys: List[str], base_state_dict: Dict[str, torch.Tensor],
```
VladVin, line 144: `generate_required_state_dict`

src/constructor/load.py
```diff
@@ -0,0 +1,220 @@
...
+        raise ValueError(f'Error in load checkpoint function. '
+                         f'Keys `{bad_keys}` in base checkpoint not in model state dict')
+
+    if override_name2ckpt_path is None:
+        override_name2state_dict = dict()
+    else:
+        override_name2state_dict = {name: load_state_dict(ckpt_path) 
+                                    for name, ckpt_path in override_name2ckpt_path.items()}
+
+    rquire_state_dict = generate_require_state_dict(model_keys, base_state_dict,
```
VladVin, line 217: `required_state_dict`

src/constructor/config_structure.py
```diff
@@ -95,6 +95,9 @@ class MetricParams:
 class TaskParams:
     name: str
     params: Optional[Dict] = field(default_factory=dict)
+    base_checkpoint: Optional[str] = None
+    override_checkpoints: Optional[Dict[str, str]] = None
+    exclude_names: Optional[List[str]] = None
```
VladVin, line 100: `exclude_keys`

src/constructor/load.py
```diff
@@ -0,0 +1,220 @@
...
+        straighten_keys = create_straighten_keys(key, model_keys)
+        depth = len(key.split('.'))
+        if len(straighten_keys) == 0:
+            raise ValueError(f'Error in load checkpoint function. '
+                             f'Key `{key}` in ovrride checkpoint not in model state dict')
+        for straighten_key in straighten_keys:
+            prefix_len = len(key) + 1
+            override_key = straighten_key[prefix_len: ]
+            # If override_key not in state dict i.e override checkpoint not contain full weights for module
+            if override_key not in state_dict:
```
VladVin, line 134: How this can happen? If the keys are straightened, then you shouldn't even try to find a part of the key in state_dict: `if override_key not in state_dict`

src/constructor/load.py
```diff
@@ -0,0 +1,220 @@
...
+    if override_name2ckpt_path is None:
+        override_name2state_dict = dict()
+    else:
+        override_name2state_dict = {name: load_state_dict(ckpt_path) 
+                                    for name, ckpt_path in override_name2ckpt_path.items()}
+
+    rquire_state_dict = generate_require_state_dict(model_keys, base_state_dict, 
+                                                    override_name2state_dict, exclude_names)
+
+    model.load_state_dict(rquire_state_dict, strict=False)
```
VladVin, line 220: Set `strict=True`

src/constructor/load.py
```diff
@@ -0,0 +1,220 @@
...
+    Contains 2 dictionary:
+        First normal state dict self.__state_dict.
+        Second dict with model key and added state dict depth key. Where the depth is the count of dots.
+    """
+    def __init__(self):
+        """Initialize structure with empty dicts."""
+        self.__state_dict = dict()
+        self.__state_dict_depth = dict()
+
+    def __add(self, key: str, weight: torch.Tensor, depth: int):
```
VladVin, line 56: `__add_or_update`

src/constructor/load.py
```diff
@@ -0,0 +1,220 @@
...
+        >>>     'backbone.linear.2': torch.tensor(3)
+        >>> }
+        >>> backbone_linear_1 = {
+        >>>     'backbone.linear.1': torch.tensor(10)
+        >>> }
+        >>> override_name2state_dict = {
+        >>>     'backbone': backbone,
+        >>>     'backbone.linear.1': backbone_linear_1
+        >>> }
+        >>> create_override_state_dict(override_name2state_dict)
```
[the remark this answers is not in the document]
VladVin: Why did you remove the example? It was good overall

src/constructor/load.py
```diff
@@ -0,0 +1,221 @@
...
+        state_dict: State dict whose keys should be prefixed.
+
+    Returns:
+        state_dict_with_prefix: Prefixed state dict.
+    """
+    state_dict_with_prefix = OrderedDict()
+    # remove spaces and dots
+    prefix = prefix.strip(' .')
+    for key, value in state_dict.items():
+        if not key.startswith(prefix):
```
VladVin, line 94: What if a key does itself contain a substring starting with a prefix? Example:
```
override_state_dict = {
  'linear': {
    'linear_layer': torch.Tensor(...)
  }
}
```
The absolute key must be: `linear.linear_layer`, but your algorithm will make it just `linear_layer`. I suggest leaving the job of putting the correct prefixes to a user because it's unclear where they start. So, just always put prefix before the key (and I guess there's no need for a special method)

src/constructor/load.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+    required_state_dict = OrderedDict()
+
+    # Check base state dict
+    check_state_dict_keys(base_state_dict, model_keys)
+
+    # Add prefix for every overridden state dict
+    overridden_full_name2state_dict = dict()
+    for overridden_name, state_dict in overridden_name2state_dict.items():
+        prefixed_state_dict = get_state_dict_with_prefix(prefix=overridden_name, state_dict=state_dict)
```
VladVin, line 153: I suggest leaving the job of putting the correct prefixes to a user because it's unclear where they start

src/constructor/load.py
```diff
@@ -0,0 +1,221 @@
...
+    # Firstly change model state dict by base state dict
+    required_state_dict.update(base_state_dict)
+    
+    # Then change model state dict with overridden state dicts, in order of it's depths
+    for depth_state_dicts in sorted_state_dicts:
+        for state_dict in depth_state_dicts:
+            required_state_dict.update(state_dict)
+
+    # Create straightened exclude keys
+    straightened_exclude_keys = []
```
VladVin, line 170: May be rename `straightened` / `straighten` with `absoulute` everywhere?

tests/constructor/test_load_checkpoint.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def test_generate_required_state_dict_when_full_parameters_was_defined(self):
+        overridden_state_dict = {
+            'layer1': {
+                'module.conv1.weight': 11,
+                'linear.weight': 22
+            }
+        }
+
+        exclude_names = ['layer1.module']
```
VladVin, line 97: `exclude_keys`

VladVin (changes requested):

src/constructor/load.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+    checkpoint = load(checkpoint_path, map_location=map_location)
+    if 'state_dict' in checkpoint:
+        state_dict = checkpoint['state_dict']
+    else:
+        state_dict = checkpoint
+    return state_dict
+
+
+def check_state_dict_keys(checked_state_dict: OrderedDict[str, torch.Tensor], model_keys: List[str],
```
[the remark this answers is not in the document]
PososikTeam: Yes we really need this. Let's explain me the case.
User put the base checkpoint with all keys, then he put overridden checkpoint with for example only half the weights.
So in this case `nn.Module.load` will not raise the error because missing half of overridden checkpoint would be get from base checkpoint. 

In fact it depend on implementation of load method, now it update the dict, but if we remove from dict all the straightened keys from overridden checkpoint it will raise by nn.Module.load.

But the next thing won't let me do it. More code in generating state dict, raise will not send you information in which checkpoint arose the error.

VladVin (changes requested):

src/constructor/load.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    depth2override_state_dicts = dict()
+
+    for override_key, override_state_dict in override_name2state_dict.items():
+        depth = len(override_key.split('.')) - 1
+        if depth not in depth2override_state_dicts:
+            depth2override_state_dicts[depth] = [override_state_dict]
+        else:
+            depth2override_state_dicts[depth].append(override_state_dict)
+    depth2override_state_dicts = OrderedDict(sorted(depth2override_state_dicts.items()))
```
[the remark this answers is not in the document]
PososikTeam: I think for dict it very bad idea so to do this need to write smth like 
```
from operator import itemgetter, attrgetter
sorted(d.items(), key=itemgetter(0)) 
```
https://docs.python.org/3/howto/sorting.html - possible variants, to write the depth word we need to create own structure.

I think comment would be enought

PR #44, opened 2022-06-14
Change classification task

VladislavPatrushev:


src/tasks/classification.py
```diff
@@ -16,25 +16,67 @@ def __init__(self, hparams: DictConfig):
...
+            NotImplementedError: if backbone, neck, pooling or head is not implemented.
         """
         super().__init__(hparams)
-        backbones_params = self._hparams.task.params.backbone_params
-        self.backbone = BACKBONES.get(self._hparams.task.params.backbone_name)(**backbones_params)
+        # BACKBONE
+        backbone_name = self._hparams.task.params.get('backbone_name')
+        backbones_params = self._hparams.task.params.get('backbone_params', dict())
+
+        backbone = BACKBONES.get(backbone_name)
```
VladVin, line 28: There will be a key error in the case of a non-existing backbone, so you don't need to check it additionally

VladVin (changes requested):

PR #47, opened 2022-06-21
SwinV2

PososikTeam:
Screenshot from 2022-07-20 11-06-55
R@1 = 84.5, for example our resnet50 have 80


src/models/backbones/swin.py
```diff
@@ -0,0 +1,451 @@
...
+            nn.init.constant_(blk.norm2.bias, 0)
+            nn.init.constant_(blk.norm2.weight, 0)
+
+
+class SwinTransformerV2(BaseModel):
+    r""" Swin Transformer V2
+        A PyTorch impl of : `Swin Transformer V2: Scaling Up Capacity and Resolution`
+            - https://arxiv.org/abs/2111.09883
+    Args:
+        img_size (int | tuple(int)): Input image size. Default 224
```
VladVin, line 144: Documentation isn't in a Google's style. The Swin official model is very well implemented, but this might be not the case with other models, so we have to restrict model implementations to our style

src/models/backbones/swin.py
```diff
@@ -0,0 +1,451 @@
+"""TorchOK Swin Transformer V2
+
+Adapted from https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/swin_transformer_v2.py
```
VladVin, line 3: How much is actually adpoted from timm? If the weights only, then I suggest to put this note above the weights configuration block

src/models/backbones/swin.py
```diff
@@ -0,0 +1,451 @@
...
+            stem=r'^absolute_pos_embed|patch_embed',  # stem and embed
+            blocks=r'^layers\.(\d+)' if coarse else [
+                (r'^layers\.(\d+).downsample', (0,)),
+                (r'^layers\.(\d+)\.\w+\.(\d+)', None),
+                (r'^norm', (99999,)),
+            ]
+        )
+
+    @torch.jit.ignore
+    def set_grad_checkpointing(self, enable=True):
```
VladVin, line 259: How is this might be used in TorchOk? Overall, check every method, it doesn't look like we use every one

src/models/backbones/swin.py
```diff
@@ -0,0 +1,451 @@
...
+        # B L C
+        x = self.feature_norms[leyer_number](x)
+        H = self.input_resolutions[leyer_number][0]
+        W = self.input_resolutions[leyer_number][1]
+        C = self.encoder_channels[leyer_number]
+        # B H W C
+        x = x.view(-1, H, W, C).permute(0, 3, 1, 2).contiguous()
+        return x
+
+    def forward_features(self, x):
```
VladVin, line 289: Put forward methods closer to the constructor

src/models/backbones/swin.py
```diff
@@ -0,0 +1,451 @@
...
+        downsample_attn = self.__forward_patch_emb(x)
+        features = []
+        for i, layer in enumerate(self.layers):
+            downsample_attn, attn = layer(downsample_attn)
```
VladVin, line 293: Why is there only one output from this layer in the officical documentation?

src/models/modules/blocks/drop.py
```diff
@@ -0,0 +1,154 @@
...
+        block_mask = 1 - block_mask
+        normalize_scale = (block_mask.numel() / block_mask.to(dtype=torch.float32).sum().add(1e-6)).to(dtype=x.dtype)
+        if inplace:
+            x.mul_(block_mask * normalize_scale)
+        else:
+            x = x * block_mask * normalize_scale
+    return x
+
+
+class DropBlock2d(nn.Module):
```
VladVin, line 93: Where is it used?

src/models/backbones/swin.py
VladVin: Keep in mind that timm has a different license than Microsoft's repo

VladVin (changes requested):

src/models/modules/blocks/drop.py
```diff
@@ -0,0 +1,154 @@
...
+            return x
+        if self.fast:
+            return drop_block_fast_2d(
+                x, self.drop_prob, self.block_size, self.gamma_scale, self.with_noise, self.inplace)
+        else:
+            return drop_block_2d(
+                x, self.drop_prob, self.block_size, self.gamma_scale, self.with_noise, self.inplace, self.batchwise)
+
+
+def drop_path(x, drop_prob: float = 0., training: bool = False, scale_by_keep: bool = True):
```
VladVin, line 126: Depending on whether DaVit gets merged first, you should take one of the two implementations

src/models/modules/bricks/mlp.py
```diff
@@ -0,0 +1,23 @@
+"""TorchOK Multi layer Perceptron module."""
+import torch
+import torch.nn as nn
+
+
+class Mlp(nn.Module):
```
VladVin, line 6: Depending on whether DaViT gets merged first, you should take one of the two implementations

VladVin (changes requested):

src/models/modules/blocks/drop.py
PososikTeam: Ok

src/models/backbones/swin.py
PososikTeam: So in feature we add only attn, not downsample_attn

src/models/backbones/swin.py
```diff
@@ -0,0 +1,456 @@
...
+            else:
+                x = blk(x)
+        attention_out = x
+        if self.downsample is not None:
+            downsample_attention = self.downsample(x)
+        else:
+            downsample_attention = x
+        return downsample_attention, attention_out
+
+    def extra_repr(self) -> str:
```
VladVin, line 129: Not used

VladVin (changes requested):

PR #48, opened 2022-06-22
feature-onnxtask

VladislavPatrushev:


src/tasks/onnx.py
```diff
@@ -0,0 +1,94 @@
...
+        super().__init__(hparams)
+        self.__infer_params = self._hparams.task.params
+        model_path = self.__infer_params.path_to_onnx
+
+        self._check_integrity_onnx_model(model_path)
+
+        self._sess = onnxrt.InferenceSession(model_path)
+        self.__binding = self._sess.io_binding()
+
+        self.__input_name = self._sess.get_inputs()[0].name
```
VladVin, line 34: Why is there one input and output only?

src/tasks/onnx.py
```diff
@@ -0,0 +1,94 @@
...
+        self.__input_shape = self._sess.get_inputs()[0].shape
+
+        self.__label_name = self._sess.get_outputs()[0].name
+        self.__output_shape = self._sess.get_outputs()[0].shape
+
+    def forward(self, x: Tensor) -> Tensor:
+        pass
+
+    def forward_with_gt(self, batch: Dict[str, Any]) -> Dict[str, Tensor]:
```
VladVin, line 43: You can skip implementing these methods

src/tasks/onnx.py
```diff
@@ -0,0 +1,94 @@
...
+        """Forward onnx model."""
+        self.__binding.bind_input(
+                name=self.__input_name,
+                device_type=self.device,
+                device_id=0,
+                element_type=np.float32,
+                shape=self.__input_shape,
+                buffer_ptr=x.data_ptr())
+
+        output_tensor = torch.empty(self.__output_shape, dtype=torch.float32, device=self.device).contiguous()
```
VladVin, line 61: The model might be of dynamic shape. So, you should derive the dynamic axes from the input tensor and the other from the output shape. Check: https://pytorch.org/docs/stable/onnx.html#torch.onnx.export

src/tasks/onnx.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self.__binding.bind_output(
+            name=self.__label_name,
+            device_type=self.device,
+            device_id=0,
```
VladVin, line 66: Why is device id hard coded?

src/tasks/onnx.py
```diff
@@ -0,0 +1,94 @@
...
+        """Complete test loop."""
+        output = self.forward_infer_with_gt(batch)
+        self._metrics_manager.forward(Phase.TEST, **output)
+
+    def predict_step(self, batch: Dict[str, Any], batch_idx: int) -> Tensor:
+        """Complete predict loop."""
+        output = self.foward_infer(batch['image'])
+        return output
+
+    def _check_integrity_onnx_model(self, model_path: str):
```
VladVin, line 92: According to the method's name, there should only be the operation of "checking", but you put the load operation as well. I suggest removing this method and doing everything inside `__init__`

VladVin (changes requested):

examples/configs/onnx_infer.yml
```diff
@@ -0,0 +1,82 @@
+# defaults:
```
VladVin, line 1: Remove

examples/configs/onnx_infer.yml
```diff
@@ -0,0 +1,82 @@
...
+#   - _self_
+#   - base_configparams
+
+task:
+  name: ONNXTask
+  params:
+    path_to_onnx: '/workdir/vpatrushev/small2/resnet18-v2-7.onnx'
```
VladVin, line 8: This is an example, don't put hard-coded paths. Also, I suggest to make this example a follow-up of the CIFAR10 training example mentioned in README. And put the command to README

examples/configs/onnx_infer.yml
```diff
@@ -0,0 +1,82 @@
...
+    providers: ['CUDAExecutionProvider']
+    inputs:
+      - shape: [1, 3, &height 224, &width 224]
+        dtype: 'float32'
+  
+
+data:
+  TEST:
+    - dataloader:
+        batch_size: 1
```
VladVin, line 19: Can we use a larger batch size? If so, make it larger. Otherwise, a user might think that we only work with `batch_size=1` which isn't ok

examples/configs/onnx_infer.yml
```diff
@@ -0,0 +1,82 @@
...
+        transform:
+          - *resize
+          - *normalize
+          - *totensor
+
+trainer:
+  accelerator: "gpu"
+  max_epochs: 1
+
+log_dir: '/workdir/logs/classification'
```
VladVin, line 66: Again, make it relative to the cache folder as it's done in another example config

src/tasks/onnx.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self._sess = onnxrt.InferenceSession(model_path, providers=self.__infer_params.providers)
+        self.__binding = self._sess.io_binding()
+
+        self.__input_names = [input_name.name for input_name in self._sess.get_inputs()]
+        self.__input_shapes = [item['shape'] for item in self.__infer_params.inputs]
+
+        dynamic_dim = self.__input_shapes[0][0]
+
+        self.__label_name = self._sess.get_outputs()[0].name
```
VladVin, line 41: The same question - why do you only process a single output?

VladVin (changes requested):

src/tasks/onnx.py
VladVin: I mean leave 'pass' here, but implement `forward`

src/tasks/onnx.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def forward_with_gt(self, batch: Dict[str, Any]) -> Dict[str, Tensor]:
+        pass
+
+    def foward_infer(self, x: Tensor) -> Tensor:
+        """Foward inference."""
+        output = self._forward_onnx(x)
+        return output
+
+    def _forward_onnx(self, inputs: Union[List[Tensor], Tensor]) -> Tensor:
```
VladVin, line 56: Why do you need a separate method? Just implement `foward_infer`

VladVin (changes requested):

VladVin (changes requested):

VladVin (changes requested):

README.md
```diff
@@ -57,7 +57,13 @@ tensorboard --logdir ~/.cache/torchok/logs/cifar10
 ### Export to ONNX
 TODO
 ### Test ONNX model
-TODO
+For test example we can use the default configuration `examples/configs/onnx_infer.yml`.
```
PososikTeam, line 60: Is it start? Because all configs extension is `yAml` anyway change it to yaml please

PososikTeam: I think it would be better if we say
`For the ONNX model test, we can use the` because we also can test ckpt also.

src/tasks/base.py
```diff
@@ -23,13 +23,17 @@ def __init__(self, hparams: DictConfig):
         self.save_hyperparameters(hparams)
         self.__constructor = Constructor(hparams)
         self._input_tensors = []
+        self._input_names = []
         self._losses = self.__constructor.configure_losses()
         self._hparams = hparams
         self._metrics_manager = self.__constructor.configure_metrics_manager()
+        
+        if hparams.task.params.get('inputs') is not None:
```
PososikTeam, line 31: need to describe what is the inputs, and why it needs, in commets

src/tasks/onnx.py
```diff
@@ -0,0 +1,107 @@
...
+                shape=(self.batch_dim, *output_params['shape'][1:]),
+                buffer_ptr=output_tensor.data_ptr())
+
+            output[output_params['name']] = output_tensor
+
+        self._sess.run_with_iobinding(self.__binding)
+        return output
+
+    def forward_infer_with_gt(self, batch: Dict[str, Any]) -> Dict[str, Tensor]:
+        """Abstract forward method for test stage."""
```
PososikTeam, line 92: It not abstarct

src/tasks/onnx.py
```diff
@@ -0,0 +1,107 @@
...
+        """Init ONNXTask.
+
+        Args:
+            hparams: Hyperparameters that set in yaml file.
+        """
+        super().__init__(hparams)
+        self.__infer_params = self._hparams.task.params
+        model_path = self.__infer_params.path_to_onnx
+
+        onnx_model = onnx.load(model_path)
```
PososikTeam, line 34: Is we need onnx.load only for checking? If yes what exactly checker do?

src/tasks/onnx.py
```diff
@@ -0,0 +1,107 @@
...
+        onnx.checker.check_model(onnx_model)
+
+        self._sess = onnxrt.InferenceSession(model_path, providers=self.__infer_params.providers)
+        self.__binding = self._sess.io_binding()
+
+        self.__inputs = [{'name': item.name,
+                          'dtype': self.str_type2numpy_type[item.type]}
+                          for item in self._sess.get_inputs()]
+
+        self.onnx2dataset = self._hparams.task.params.keys_mapping_onnx2dataset
```
PososikTeam, line 44: why it not private? I think it's better to change name to keys_mapping_onnx2dataset like in yaml

src/tasks/onnx.py
```diff
@@ -0,0 +1,107 @@
...
+                           'shape': item.shape,
+                           'dtype': self.str_type2numpy_type[item.type]} for item in self._sess.get_outputs()]
+
+    def forward(self, x: Tensor) -> Tensor:
+        pass
+
+    def forward_with_gt(self, batch: Dict[str, Any]) -> Dict[str, Tensor]:
+        pass
+
+    def foward_infer(self, inputs: Dict[str, Tensor]) -> Tensor:
```
PososikTeam, line 56: Return type is Tensor, but it actually Dict

src/tasks/onnx.py
```diff
@@ -0,0 +1,107 @@
...
+            output[output_params['name']] = output_tensor
+
+        self._sess.run_with_iobinding(self.__binding)
+        return output
+
+    def forward_infer_with_gt(self, batch: Dict[str, Any]) -> Dict[str, Tensor]:
+        """Abstract forward method for test stage."""
+        input_data = batch['image']
+        target = batch['target']
+        prediction = self.foward_infer(input_data)
```
PososikTeam, line 95: So it would be work if 
`batch_image = batch['image']['input_1']`?
Because `forward_infer` input is dict.

You need define dataset fro run this task, i think it might be `ImageClassificationDataset`. And `foward_infer` must support 2 inputs Tensor and Dict[str, Tensor]. Also write dataset type in Task documentation for that task will be work.

torchok/__main__.py
```diff
@@ -30,6 +30,5 @@ def entrypoint(config: DictConfig):
     trainer = create_trainer(config)
     trainer.fit(model, ckpt_path=config.resume_path)
 
-
```
PososikTeam, line 33: Need 2 space line, please return this empty line

torchok/tasks/onnx.py
```diff
@@ -0,0 +1,111 @@
...
+        logging.info(f'Input onnx names: {input_names}')
+
+        self.__keys_mapping_onnx2dataset = self._hparams.task.params.keys_mapping_onnx2dataset
+
+        self.__outputs = [{'name': item.name,
+                           'shape': item.shape,
+                           'dtype': self.str_type2numpy_type[item.type]} for item in self._sess.get_outputs()]
+
+        output_names = [output['name'] for output in self.__outputs]
+        logging.info(f'Output onnx names: {output_names}')
```
PososikTeam, line 53: Output onnx names -> ONNX model output names

PR #49, opened 2022-06-22
featue save_onnx_ckpt

VladislavPatrushev:


examples/configs/classification_cifar10.yaml
```diff
@@ -85,6 +85,7 @@ checkpoint:
   monitor: valid/F1Score
   save_last: true
   mode: max
+  onnx_to_save: true
```
VladVin, line 88: Better to call `export_to_onnx` because the operation does exactly the export thing

VladVin (changes requested):

PR #53, opened 2022-06-29
SOP training

PososikTeam:
Solve issue with metric manager,  it not reset values after epoch end. Also overwrite classification retrieval metric. Add HitRate@k metric, use it as R@1 as show in https://paperswithcode.com/sota/metric-learning-on-stanford-online-products-1 leaderboard SOP, (HitRate@1 = R@1 for this metrics).

So, i reach 80% R@1 with Arcface and ResNet50, with 512 embedding size
Screenshot from 2022-06-29 08-20-38
Screenshot from 2022-06-29 08-20-51
.

Also write Linear Pooling and Eval notebook.

Graphics after review:
Screenshot from 2022-07-05 10-10-01
Screenshot from 2022-07-05 10-09-52


src/metrics/representation.py
```diff
@@ -340,11 +315,16 @@ def generator():
                                [6, 1]])
                     ].
             """
-            for i in range(0, len(queries), self.search_batch_size):
-                query_idxs = np.arange(i, min(i + self.search_batch_size, len(queries)))
+            for i in range(0, len(queries_idxs), self.search_batch_size):
+                idxs = np.arange(i, min(i + self.search_batch_size, len(queries_idxs)))
+
+                closest_dist, closest_idx = index.search(vectors[idxs], k=k)
```
VladVin, line 321: Shouldn't you access query vectors like this?  
`vectors[queries_idxs[idxs]]`

src/metrics/representation.py
```diff
@@ -340,11 +315,16 @@ def generator():
...
+                # remove first element which is actually is it's classification dataset
```
VladVin, line 322: Remove the first retrieved element in case of classification dataset since it's a query itself

src/metrics/representation.py
```diff
@@ -357,15 +337,15 @@ def generator():
                     np.arange(len(closest_idx)))
                 
                 if scores is None:
-                    relevant = map(lambda r: np.stack((r, np.ones_like(r)), axis=1), relevant)
+                    batch_relevant = map(lambda r: np.stack((r, np.ones_like(r)), axis=1), batch_relevant)
                 else:
-                    relevant = map(lambda r_q: np.stack((r_q[0], scores[r_q[0], q_order_idxs[r_q[1]]]), axis=1), \
-                        zip(relevant, query_idxs))
+                    batch_relevant = map(lambda r_q: np.stack((r_q[0], scores[r_q[0], queries_idxs[r_q[1]]]), axis=1), \
+                        zip(batch_relevant, idxs))
```
VladVin, line 343: Rename `idxs` -> `batch_idxs`

src/metrics/representation.py
```diff
@@ -252,43 +242,24 @@ def prepare_classification_data(self, vectors: np.ndarray, targets: np.ndarray
...
-        relevant = relevant[correct_classes]
-        query_idxs = query_idxs[correct_classes]
-
-        # Retrieve query and database vectors and build index base on latter.
-        q_vecs = vectors[query_idxs]
-        db_vecs = vectors[db_idxs]
+        target_values, target_counts = np.unique(targets, return_counts=True)
+        relevants = []
+        for target, count in zip(target_values, target_counts):
+            relevants += [np.where(targets == target)[0]] * count
```
VladVin, line 253: Good implementation, but put a comment explaining why you are doing like this

VladVin (changes requested):

src/data/datasets/examples/sop.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return sample
+
+    def __len__(self) -> int:
+        """Dataset length."""
+        return len(self.__csv)
+
+    def _check_integrity(self) -> bool:
+        """Check integrity."""
+        if os.path.exists(self.__path):
```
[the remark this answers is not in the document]
PososikTeam: In CIFAR10 it use `from torchvision.datasets.utils import check_integrity` with the next code
```
def check_integrity(fpath: str, md5: Optional[str] = None) -> bool:
    if not os.path.isfile(fpath):
        return False
    if md5 is None:
        return True
    return check_md5(fpath, md5)```
    
 If i need to change it?

File: src/metrics/metrics_manager.py [hunk 6]
```diff
@@ -41,7 +41,9 @@ def forward(self, *args, **kwargs):
         return self.__metric(*args, **kwargs)
 
     def compute(self):
-        return self.__metric.compute()
+        value = self.__metric.compute()
+        self.__metric.reset()
```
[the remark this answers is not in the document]
PososikTeam: Ok it actually bad idea to call reset in MetricWithUtils it is better to call it in Metric implementation classes. So i call reset in the end of IndexBasedMeter in compute method.

src/metrics/representation.py
```diff
@@ -146,228 +150,225 @@ def compute(self) -> float:
...
 
-        for idx in range(len(q_vecs)):
-            relevant_idxs = np.where(scores[:, queries_idxs[idx]] > 0.)[0]
+        Raises:
+            ValueError: If dataset has query vector without relevants.
+        """
+        is_queries = query_order_numbers >= 0
+        full_indexes = np.arange(len(query_order_numbers))
+        clear_query_order_numbers = query_order_numbers[is_queries]
+        queries_idxs = full_indexes[is_queries]
```
VladVin, line 242: Looks like `queries_idxs` is exactly equal to `query_order_numbers`

src/metrics/representation.py
```diff
@@ -146,228 +150,225 @@ def compute(self) -> float:
...
-
-        # Group item indices of the same class
         groups = pd.Series(ts.groupby(ts).groups)
+        # now groups contain group indexes with one targets
+        relevants = []
+        queries_idxs = []
+        # create queries with its relevants indexes
+        for i, group in enumerate(groups):
+            for query_idx in group:
+                # need to drop from relevants index which equal query index
```
VladVin, line 294: Why did you change the implementation back to pandas version? It was good IMHO

VladVin (changes requested):

src/metrics/metrics_manager.py
```diff
@@ -23,25 +20,37 @@ def __init__(self, metric: Metric, mapping: Dict[str, str], log_name: str):
...
     @property
     def mapping(self) -> Dict[str, str]:
         """Dictionary for mapping Metric forward input keys with Task output dictionary keys."""
-        return self.__mapping
-
+        return self._mapping
+    
     def forward(self, *args, **kwargs):
-        return self.__metric(*args, **kwargs)
+        """Update metric states."""
```
VladVin, line 43: Is documentation still needed? Also, do we still need forward if it calls `update`?

VladVin (changes requested):

src/metrics/representation.py
PososikTeam: it's equal only when dataset is has attribute `shuffle=False`

src/metrics/metrics_manager.py
PososikTeam: Yes, we need forward because it's nn.Module

src/metrics/representation.py
PososikTeam: it was bad implementation, because for `shuffle=True` it won't be work. It was only works when class arranged in order like [0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3 ....]. If not use pandas, we need to create groups like in this implementation, so we need call np.uniq then np.where len(np.uniq) times and only then create cycle for delete query element.

VladVin (changes requested):

src/metrics/representation.py
```diff
@@ -146,228 +153,243 @@ def compute(self) -> float:
...
             # if classification dataset
             targets = torch.cat(self.targets).numpy()
             # prepare data
-            q_vecs, db_vecs, relevants, scores, \
-                db_idxs, q_order_idxs = self.prepare_classification_data(vectors, targets)
+            relevant_idxs, gallery_idxs, query_row_idxs = self.prepare_classification_data(targets)
+            # mock scores and query column indexes because it belong to representation data
+            scores = None
+            query_column_idxs = None
+            # all query vectors is in relevant, so create array with True elements
```
VladVin, line 170: I've seen a row `relevant = group.drop(query_idx).tolist()` in `prepare_classification_data`. Doesn't it mean that you removed the queries from relevant already?

src/metrics/representation.py
```diff
@@ -146,228 +153,243 @@ def compute(self) -> float:
...
+            query_as_relevant = np.array([True] * len(gallery_idxs))
```
VladVin, line 171: Not consistent with the `prepare_representation_data` call where you return `query_as_relevant` array - I think you should do the same way with `prepare_classification_data`

src/metrics/representation.py
```diff
@@ -146,228 +153,243 @@ def compute(self) -> float:
...
+        Raises:
+            ValueError: If dataset has query vector without relevants.
+        """
+        full_indexes = np.arange(len(query_idxs))
+        is_query = query_idxs >= 0
+        # array of columns indexes in scores matrix, need to get relevant
+        query_column_idxs = query_idxs[is_query]
+        # array of row indexes in scores matrix, also this index belong to vectors storage, need to get query vector in
+        # search study
+        query_row_idxs = full_indexes[is_query]
```
VladVin, line 252: Actually, this is `np.where(is_query)[0]`. You don't need `np.arange`

src/metrics/representation.py
```diff
@@ -146,228 +153,243 @@ def compute(self) -> float:
...
+        # gallery idxs in vectors storage, which row sum > 0
+        gallery_idxs = np.where(np.sum(np.abs(scores), axis=-1) > 0.)[0]
```
VladVin, line 254: I think this isn't a problem if you store queries in relevant. On search time you can exclude queries since they are always at top-1. The implementation will be simpler and equivalent for both types of datasets. The second argument is that doing this way `gallery_idxs` might exclude queries that are themselves relevant to other queries

VladVin (changes requested):

src/metrics/representation.py
PososikTeam: In classification data i remove query from relevants, not in gallery https://github.com/eora-ai/torchok/pull/53/files#:~:text=ts%20%3D%20pd.Series,relevant_idxs.append(relevant). 
In  representation dataset i not remove query from relevant because the score matrix should contain zeros in diagonal elements (but it's not checked). If i need to check it? 
If you ask why i remove from relevant, i told you because many metric after compute divide metric to `len(relevants)`

src/metrics/representation.py
PososikTeam: Actually i really don't know what will be faster compute `np.where(np.sum(np.abs(scores), axis=-1) > 0.)[0]` this operation or add query to gallery and search. I think it would be search because we do search many times. I think it's good idea to get this parameter from dataset to not compute this operation.

src/metrics/representation.py
```diff
@@ -125,249 +132,263 @@ def update(self, vectors: torch.Tensor, targets: Optional[torch.Tensor] = None,
...
+                if len(relevant) == 0:
+                    raise ValueError(f'Representation metric. The class {groups.index[i]} has only one element.')
+                query_row_idxs.append(query_idx)
+                relevant_idxs.append(relevant)
+
+        relevant_idxs = np.array(relevant_idxs)
+        query_row_idxs = np.array(query_row_idxs)
+        gallery_idxs = np.arange(len(targets))
+        # all query vectors is in relevant, so create array with True elements
+        query_as_relevant = np.array([True] * len(gallery_idxs))
```
VladVin, line 308: Much faster version: `query_as_relevant = np.full((len(gallery_idxs),), fill_value=True, dtype=np.bool)`

VladVin (changes requested):

src/metrics/representation.py
VladVin: There's also a way to calculate `np.any` by some axis

PR #54, opened 2022-06-30
MLFlow Logger

PososikTeam:
add mlflow logger, with save artefacts after exception, also add create_datetime_log_subdir parameter


src/constructor/logger.py
```diff
@@ -0,0 +1,116 @@
...
+        tags=logger_params.tags,
+        save_dir=str(save_dir),
+        artifact_location=str(remote_save_dir)
+    )
+
+    return logger
+
+
+def create_logger(logger_params, outputs_path, experiment_name, experiment_subdir, job_link):
+    if logger_params.logger == LoggerType.TENSORBOARD:
```
VladVin, line 112: There shouldn't be any ad-hocs related to specific loggers

VladVin (changes requested):

src/constructor/logger.py
```diff
@@ -0,0 +1,171 @@
...
+    if create_datetime_log_subdir:
+        experiment_subdir = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    else:
+        experiment_subdir = ''
+
+    log_dir = Path(log_dir)
+    full_outputs_path = log_dir / experiment_name / experiment_subdir
+    full_outputs_path.mkdir(exist_ok=True, parents=True)
+
+    return experiment_subdir, full_outputs_path
```
VladVin, line 37: You don't need to return `experiment_subdir`. It can easily be obtained via `str(full_outputs_path.relative_to(log_dir/experiment_name))`

src/constructor/logger.py
```diff
@@ -0,0 +1,171 @@
...
+                         tags=tags, save_dir=save_dir, prefix=prefix, artifact_location=artifact_location)
+        self._save_dir = Path(save_dir)
+
+    @rank_zero_only
+    def finalize(self, status: str = 'FINISHED') -> None:
+        """Call finalize of pytorch lightning MlFlowLogger and logs *.ckpt and *.onnx artifacts in artifact_location."""
+        super(mlflow.MLFlowLogger, self).finalize(status)
+        status = 'FINISHED' if status == 'success' else status
+
+        base_dir = str(self._save_dir)
```
VladVin, line 100: Why do you make a rename? Also, the `pathlib` also contains glob operations

src/constructor/logger.py
```diff
@@ -0,0 +1,171 @@
...
+        upload_file_paths = glob.glob(base_dir + '/*.ckpt') + glob.glob(base_dir + '/*.onnx')
```
VladVin, line 101: Very bad way of paths concatenating. Won't work on Windows

src/constructor/logger.py
```diff
@@ -0,0 +1,171 @@
...
+        experiment_subdir = ''
+
+    log_dir = Path(log_dir)
+    full_outputs_path = log_dir / experiment_name / experiment_subdir
+    full_outputs_path.mkdir(exist_ok=True, parents=True)
+
+    return experiment_subdir, full_outputs_path
+
+
+class MLFlowLogger(mlflow.MLFlowLogger):
```
VladVin, line 40: Rename the class to `MLFlowLoggerX` and allow a user to choose between both: our version and Lighting's one

VladVin (changes requested):

environment.yml
```diff
@@ -16,18 +15,6 @@ dependencies:
   - cudatoolkit=11.1
   - pytorch-lightning=1.6.2
   - torchmetrics=0.9.2
-=======
```
VladVin, line 19: Brrr

VladVin (changes requested):

PR #56, opened 2022-07-01
Feature DaViT

VladislavPatrushev:


src/models/modules/bricks/droppath.py
```diff
@@ -0,0 +1,27 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+def drop_path(x, drop_prob: float = 0., training: bool = False, scale_by_keep: bool = True):
```
VladVin, line 6: Depending on whether Swin gets merged first, you should take one of the two implementations

src/models/modules/bricks/mlp.py
```diff
@@ -0,0 +1,35 @@
+from typing import Optional
+
+import torch.nn as nn
+from torch import Tensor
+
+
+class Mlp(nn.Module):
```
VladVin, line 7: Depending on whether Swin gets merged first, you should take one of the two implementations

VladVin (changes requested):

PR #57, opened 2022-07-01
BaseModel class

PososikTeam:


src/models/base.py
```diff
@@ -213,22 +17,4 @@ def get_forward_output_channels(self) -> Union[int, List[int]]:
...
-        return forward_channels, hooks_channels
-
-    @property
-    def features_info(self):
-        return self.__features_info
-
-    @property
-    def feature_hooks(self):
-        return self.__feature_hooks
+
```
VladVin, line 20: Add abstract method `no_weight_decay`, look for the signature in `Constructor`. And remove the `hasattr` check there

VladVin (changes requested):

src/models/backbones/base_backbone.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @abstractmethod
+    def forward_features(self, *args, **kwargs):
+        """Forward method for getting backbone feature maps.
+           They are mainly used for segmentation and detection tasks.
+        """
+        pass
+
+    @abstractmethod
+    def get_forward_feature_output_channels(self) -> Union[int, List[int]]:
```
VladVin, line 23: Maybe give a shorter name `get_forward_features_channels`?

VladVin (changes requested):

src/tasks/base.py
```diff
@@ -41,9 +41,17 @@ def forward_with_gt(batch: Dict[str, Any]) -> Dict[str, torch.Tensor]:
         """Abstract forward method for training(with ground truth labels)."""
         pass
 
+    @abstractmethod
+    def no_weight_decay(self) -> List[str]:
```
VladVin, line 45: It should be not here but in `BaseModel`. Otherwise, a user has to setup no-weight-decay params in the task, but they are defined for each model custom which is overkill

VladVin (changes requested):

src/models/backbones/base_backbone.py
```diff
@@ -0,0 +1,35 @@
+from abc import ABC, abstractmethod
+from typing import List, Union, Optional
+
+from src.models.base import BaseModel
+
+
+class BaseBackbone(BaseModel, ABC):
+    """Base model for TorchOk Backbones"""
+    def __init__(self,
+                 in_channels: Optional[Union[int, List[int]]] = None,
```
VladVin, line 10: You can add `Tuple[int, ...]` into `Union` because it's also acceptable

src/models/backbones/base_backbone.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @abstractmethod
+    def forward_features(self, *args, **kwargs):
+        """Forward method for getting backbone feature maps.
+           They are mainly used for segmentation and detection tasks.
+        """
+        pass
+
+    @property
+    def out_feature_channels(self):
```
VladVin, line 31: No typings

src/models/backbones/resnet.py
```diff
@@ -143,17 +139,25 @@ def forward(self, x: torch.Tensor):
...
-        return features_info
-
-    def get_forward_output_channels(self) -> Union[int, List[int]]:
-        """Return number of output channels."""
-        return self.num_features
+    def forward_features(self, x: torch.Tensor):
+        """Forward method."""
+        features = [x]
+        x = self.convbnact(x)
+        x = self.maxpool(x)
```
VladVin, line 146: Shouldn't we also return the feature map after the "stem" layer?

src/models/backbones/resnet.py
```diff
@@ -143,17 +139,25 @@ def forward(self, x: torch.Tensor):
...
-        features_info = []
-        for num_channels, module_name, stride in zip(output_channels, module_names, strides):
-            feature_info = FeatureInfo(module_name=module_name, num_channels=num_channels, stride=stride)
-            features_info.append(feature_info)
-        return features_info
-
-    def get_forward_output_channels(self) -> Union[int, List[int]]:
-        """Return number of output channels."""
-        return self.num_features
+    def forward_features(self, x: torch.Tensor):
```
VladVin, line 142: No typings

src/models/base.py
```diff
@@ -1,234 +1,46 @@
...
-    
-    def get_output_channels(self) -> Tuple[Union[int, List[int]], List[int]]:
-        """Create forward an hooks channels numbers.
-        
-        Returns:
-            forward_channels: Forward pass output channels.
-            hooks_channels: Hooks output channels.
+
+    def no_weight_decay(self) -> List[str]:
+        """Create module names for which weights decay will not be used.
```
VladVin, line 28:
```suggestion
        """Create module names for which weight decay will not be used.
```

src/models/identity.py
```diff
@@ -1,17 +0,0 @@
-from torch import nn, Tensor
-
-from src.models.base import BaseModel
-
-
-class Identity(BaseModel):
```
VladVin, line 6: Why did you remove an Identity module? It should be in the backbones, heads and pooling registries

src/tasks/base.py
```diff
@@ -42,9 +42,19 @@ def forward_with_gt(batch: Dict[str, Any]) -> Dict[str, torch.Tensor]:
         """Abstract forward method for training(with ground truth labels)."""
         pass
 
+    def train_modules(self) -> List[str]:
+        """Create a list of modules that need to train. Method used to apply optimizer for returned modules.
```
VladVin, line 46:
```suggestion
        """Create a list of 1st-level modules that need to be optimized. The method is used to apply an optimizer for the returned modules.
```

src/tasks/base.py
```diff
@@ -42,9 +42,19 @@ def forward_with_gt(batch: Dict[str, Any]) -> Dict[str, torch.Tensor]:
...
+
+        By default, it would be self.children().
+        
+        Returns: List of modules that need to train.
```
VladVin, line 50:
```suggestion
        Returns: List of modules that need to be optimized.
```

VladVin (changes requested):

src/models/identity.py
PososikTeam: Because we decide not to use Identity in tasks, fro example in classification task we use
`neck_name = self._hparams.task.params.get('neck_name')` for get neck name but we can use
`neck_name = self._hparams.task.params.get('neck_name', 'Identity')`. So why we need this modules?

src/models/backbones/resnet.py
PososikTeam: In previous version of TorchOk it not returned, but i think we should to give flexibility

src/models/base.py
```diff
@@ -1,234 +1,50 @@
...
         """
-        forward_channels = self.get_forward_output_channels()
-        hooks_channels = self._get_hooks_output_channels()
-        return forward_channels, hooks_channels
+        return list()
 
     @property
-    def features_info(self):
-        return self.__features_info
+    def in_channels(self) -> Union[int, List[int]]:
```
VladVin, line 35: `Optional[Union[int, List[int], Tuple[int, ...]]]`

VladVin (changes requested):

environment.yml
```diff
@@ -16,18 +15,6 @@ dependencies:
   - cudatoolkit=11.1
   - pytorch-lightning=1.6.2
   - torchmetrics=0.9.2
-=======
-  - mlflow=1.22.0
-  - pandas=1.4.3
-  - pillow=9.1.1
-  - pip=22.1.2
-  - python=3.9
-  - pytorch=1.12
```
VladVin, line 25: Merge dependencies correctly

VladVin (changes requested):

PR #71, opened 2022-07-18
Segmentation

PososikTeam:


src/losses/segmentation/dice.py
```diff
@@ -0,0 +1,192 @@
...
+
+    Raises:
+        ValueError: If x not in torch.Tensor, np.ndarray, list, tuple types.
+    """
+    if isinstance(x, torch.Tensor):
+        if dtype is not None:
+            x = x.type(dtype)
+        return x
+    if isinstance(x, np.ndarray):
+        x = torch.from_numpy(x)
```
VladVin, line 76: We don't work with numpy tensors, I suggest to remove this option

VladVin (changes requested):

PR #72, opened 2022-07-20
hotfix build model with config

VladislavPatrushev:


Animatory:
Close due to alternate implementation

PR #107, opened 2022-08-08
Global refactor

Animatory:


torchok/data/datasets/base.py
```diff
@@ -28,12 +28,11 @@ def __init__(self,
             grayscale: If True, image will be read as grayscale otherwise as RGB.
             test_mode: If True, only image without labels will be returned.
         """
-        self._test_mode = test_mode
```
PososikTeam, line 31: Why remove protected, i think it must be protected, not public

PR #112, opened 2022-08-18
Feature github actions

VladislavPatrushev:


VladislavPatrushev:
Also need set setting like here https://stackoverflow.com/questions/58654530/how-to-auto-reject-a-pull-request-if-tests-are-failing-github-actions

PR #131, opened 2022-09-08
implement bce loss

PososikTeam:


torchok/losses/classification/binary_cross_entropy.py
```diff
@@ -0,0 +1,48 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+
+from typing import Union
+
+from torchok.constructor import LOSSES
+
+
+@LOSSES.register_class
+class BCEWithLogitsLossX(nn.BCEWithLogitsLoss):
```
Animatory, line 13: Rename to BCEWithLogitsLoss as it extends the original BCE without changing other functionality

PR #147, opened 2022-09-19
Feature Object detection

Animatory:


environment.yml
```diff
@@ -24,5 +24,8 @@ dependencies:
       - hydra-core==1.2.0
       - opencv-python==4.6.0.*
       - ranx==0.2.8
-      - timm==0.6.5
-      - pycocotools==2.0.4
```
PososikTeam, line 28: Is mmdet has pycocotools?

torchok/models/heads/detection/yolov3.py
```diff
@@ -0,0 +1,231 @@
...
+                scores,
+                cfg['score_thr'],
+                dict(cfg['nms']),
+                cfg['max_per_img'],
+                score_factors=objectness)
+            det_results.append(dict(bboxes=det_bboxes, labels=det_labels))
+        return det_results
+
+    @force_fp32(apply_to=('pred_maps',))
+    def loss(self, joint_loss: JointLoss, pred_maps: List[Tensor], gt_bboxes: List[Tensor], gt_labels: List[Tensor],
```
PososikTeam, line 147: In MMDetection loss and get_bboxes method has already writen. Why we need to overwrite this method here?

environment.yml
Animatory: Yes

torchok/models/heads/detection/yolov3.py
Animatory: Because we have JointLoss that cannot be adapted otherwise from mmdet

torchok/metrics/detection.py
```diff
@@ -0,0 +1,63 @@
...
+import numpy as np
+from mmdet.core.evaluation.mean_ap import eval_map
+from torch import Tensor
+from torchmetrics import Metric
+
+from torchok.constructor import METRICS
+
+
+@METRICS.register_class
+class MMDetectionMAP(Metric):
```
PososikTeam, line 12: Has it problems with memory?

torchok/models/backbones/__init__.py
```diff
@@ -6,3 +8,7 @@
 import torchok.models.backbones.mobilenetv3
 import torchok.models.backbones.resnet
 import torchok.models.backbones.swin
+
+has_mmcv = importlib.util.find_spec("mmcv")
```
PososikTeam, line 12: Why we need this? I think just import is ok

torchok/models/heads/detection/__init__.py
```diff
@@ -0,0 +1,5 @@
+import importlib
+
+has_mmcv = importlib.util.find_spec("mmcv")
```
PososikTeam, line 3: I think just import is ok

torchok/tasks/detection.py
```diff
@@ -0,0 +1,106 @@
...
+        output = self.forward_with_gt(batch)
+        output['prediction'] = self.bbox_head.get_bboxes(**output)
+        output['target'] = [dict(bboxes=bb, labels=la) for bb, la in zip(output['gt_bboxes'], output['gt_labels'])]
+        self.metrics_manager.update(Phase.TEST, **output)
+
+    def predict_step(self, batch: Dict[str, Union[torch.Tensor, int]], batch_idx: int) -> Dict[str, torch.Tensor]:
+        """Complete predict loop."""
+        output = self.forward_with_gt(batch)
+        output['prediction'] = self.bbox_head.get_bboxes(**output)
+        output['target'] = [dict(bboxes=bb, labels=la) for bb, la in zip(output['gt_bboxes'], output['gt_labels'])]
```
PososikTeam, line 105: Need to check if output has gt_bboxes and gt_labels or remove this line, because it's predict

torchok/tasks/detection.py
```diff
@@ -0,0 +1,106 @@
...
+        Args:
+            hparams: Hyperparameters that set in yaml file.
+        """
+        super().__init__(hparams)
+
+        # BACKBONE
+        backbone_name = self._hparams.task.params.get('backbone_name')
+        backbones_params = self._hparams.task.params.get('backbone_params', dict())
+        self.backbone = BACKBONES.get(backbone_name)(**backbones_params)
+        self.num_scales = self._hparams.task.params.get('num_scales', len(self.backbone.out_encoder_channels))
```
PososikTeam, line 27: Need comment that is num_scales, and why it can be less than backbone features

PososikTeam: It's also bad that we can't write parameter like `self.num_scales` in documentation, need to discuss on torchok weekly ***

torchok/tasks/detection.py
```diff
@@ -0,0 +1,106 @@
...
+        x = self.backbone.forward_features(x)[-self.num_scales:]
+        with torch.cuda.amp.autocast(dtype=torch.float32):
+            x = self.neck(x)
+        x = self.bbox_head(x)
+        return x
+
+    def forward_with_gt(self, batch: Dict[str, Union[torch.Tensor, int]]) -> Dict[str, Any]:
+        """Forward with ground truth labels."""
+        input_data = batch.get('image')
+        features = self.backbone.forward_features(input_data)[-self.num_scales:]
```
PososikTeam, line 52: Our forward feature reurn [x] + features, is mmdet also need x?

torchok/models/heads/detection/fcos.py
```diff
@@ -0,0 +1,218 @@
...
+
+        for k, v in kwargs.items():
+            if isinstance(v, DictConfig):
+                kwargs[k] = ConfigDict(OmegaConf.to_container(v, resolve=True))
+        AnchorFreeHead.__init__(self, num_classes, in_channels, **kwargs)
+
+        self.init_weights()
+
+    @staticmethod
+    def format_dict(head_output):
```
PososikTeam, line 105: may be replace this method to task?

torchok/models/heads/detection/fcos.py
```diff
@@ -0,0 +1,218 @@
...
+        num_pos = max(reduce_mean(num_pos), 1.0)
+
+        pos_bbox_preds = flatten_bbox_preds[pos_inds]
+        pos_centerness = flatten_centerness[pos_inds]
+        pos_bbox_targets = flatten_bbox_targets[pos_inds]
+        pos_centerness_targets = self.centerness_target(pos_bbox_targets)
+        # centerness weighted iou loss
+        centerness_denorm = max(reduce_mean(pos_centerness_targets.sum().detach()), 1e-6)
+
+        # if len(pos_inds) > 0:
```
PososikTeam, line 173: MMDet code in this section is
```
        if len(pos_inds) > 0:
            pos_points = flatten_points[pos_inds]
            pos_decoded_bbox_preds = self.bbox_coder.decode(
                pos_points, pos_bbox_preds)
            pos_decoded_target_preds = self.bbox_coder.decode(
                pos_points, pos_bbox_targets)
            loss_bbox = self.loss_bbox(
                pos_decoded_bbox_preds,
                pos_decoded_target_preds,
                weight=pos_centerness_targets,
                avg_factor=centerness_denorm)
            loss_centerness = self.loss_centerness(
                pos_centerness, pos_centerness_targets, avg_factor=num_pos)
        else:
            loss_bbox = pos_bbox_preds.sum()
            loss_centerness = pos_centerness.sum()
```
So the question is can losses calculate loss for empty tensors as inputs?

PososikTeam: If so remove comment

torchok/models/heads/detection/fcos.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Returns:
+            list[list[Tensor, Tensor]]: Each item in result_list is 2-tuple.
+                The first item is an (n, 5) tensor, where the first 4 columns
+                are bounding box positions (tl_x, tl_y, br_x, br_y) and the
+                5-th column is a score between 0 and 1. The second item is a
+                (n,) tensor where each item is the predicted class label of
+                the corresponding box.
+        """
+        pseudo_meta = [dict(img_shape=image_shape, scale_factor=1) for i in range(len(cls_scores[0]))]
```
PososikTeam, line 214: May you explain me why `get_bboxes` work in mmdet fcos? because default parameter for `img_metas` is None and they have a code line 
```for img_id in range(len(img_metas)):``` so it must fall.

File: examples/notebooks/CheckTransforms.ipynb [hunk 12]
```diff
@@ -3,11 +3,7 @@
   {
```
PososikTeam, line 3: Please remember me, how you solve issue with bboxes out of image shape

PososikTeam (changes requested):
Please, answer my questions and think about suggestions. Also i think you can change readme, because a lot of things in roadmap has already done

examples/notebooks/CheckTransforms.ipynb
Animatory: Check function `DetectionDataset.filter_bboxes' in torchok/data/datasets/detection/detection.py

torchok/metrics/detection.py
Animatory: The problem with memory is in another metric `CocoEvalMAP`. But this problem may be reduced by using `compute_on_step: false` parameter

torchok/models/backbones/__init__.py
Animatory: It is for cases when you do not install mmcv. Without it, CI/CD will fail

torchok/tasks/detection.py
Animatory: Fixed

torchok/tasks/detection.py
Animatory: Agree

torchok/tasks/detection.py
Animatory: No, here we take the features from the latest backbone stages like `[layer2, layer3, layer4]`

torchok/models/heads/detection/fcos.py
Animatory: No, this method is special for each detection head.

torchok/models/heads/detection/fcos.py
```diff
@@ -0,0 +1,218 @@
...
+                 num_classes,
+                 in_channels,
+                 regress_ranges=((-1, 64), (64, 128), (128, 256), (256, 512), (512, INF)),
+                 center_sampling=False,
+                 center_sample_radius=1.5,
+                 norm_on_bbox=False,
+                 centerness_on_reg=False,
+                 norm_cfg=dict(type='GN', num_groups=32, requires_grad=True),
+                 conv_cfg=None,
+                 train_cfg=None,
```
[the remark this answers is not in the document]
Animatory: These fields are taken from MMDet and are set up inside the MMDet code. I will think about how to add documentation for that

torchok/models/heads/detection/fcos.py
Animatory: In mmdet this function cannot work without `img_metas`. It is a trick that the default value is None. But inside their code the meta is used only to get image shape. Therefore in my implementation I initialize it with base image shape.

torchok/models/heads/detection/fcos.py
Animatory: It works with `len(pos_inds) == 0`

PososikTeam (approved):

PR #164, opened 2022-11-21
Fix bag with metric and add ability to choice if use batch searching or no

PososikTeam:


torchok/metrics/representation_torchmetrics.py
```diff
@@ -26,22 +26,26 @@ class TorchMetricBaseMetr(IndexBasedMeter):
...
 
     def process_data_for_metric_func(self, closest_scores: np.ndarray, closest_idxs: np.ndarray,
                                      relevants_idxs: np.ndarray, query_col_idxs: np.ndarray,
                                      scores: np.ndarray, k: int) -> List:
         preds = torch.tensor(closest_scores)
         target = torch.tensor([np.isin(closest_idxs[i], relevants_idxs[i]) for i in range(len(closest_idxs))],
                               dtype=torch.long)
-        indexes = torch.tensor([target.shape[1] * [i] for i in range(len(target))], dtype=torch.long)
+        if self.use_batching_search:
+            indexes = torch.tensor([target.shape[1] * [i] for i in range(len(target))], dtype=torch.long)
```
Animatory, line 46:
```
tar = torch.rand(500, 1024)
%timeit torch.tile(torch.arange(len(tar), dtype=torch.long)[:, None], (1, tar.shape[1]))
1.07 ms � 17.8 ns per loop (mean � std. dev. of 7 runs, 1,000 loops each)

%timeit torch.tensor([tar.shape[1] * [i] for i in range(len(tar))], dtype=torch.long)
11.1 ms � 466 ns per loop (mean � std. dev. of 7 runs, 100 loops each)
```

Animatory (approved):

PR #183, opened 2023-01-16
Add DETR

Animatory:


torchok/models/heads/detection/fcos.py
```diff
@@ -64,6 +64,7 @@ class FCOSHead(fcos_head.FCOSHead):
     """  # noqa: E501
 
     def __init__(self,
+                 joint_loss,
```
PososikTeam, line 67: why we need this?

torchok/tasks/detection.py
```diff
@@ -75,7 +80,7 @@ def forward(self, x: torch.Tensor) -> List[Dict[str, torch.Tensor]]:
         features = self.neck(features)
         features = self.bbox_head(features)
         output = self.bbox_head.format_dict(features)
-        output = self.bbox_head.get_bboxes(**output, image_shape=x.shape[-2:])
+        output = self.bbox_head.get_bboxes(**output, img_metas=[dict(img_shape=x.shape[-2:])] * x.shape[0])
```
PososikTeam, line 83: is it ok that we send dict of tuple, not dict of tensors in output?

PososikTeam: And actually it is a list

torchok/tasks/detection.py
```diff
@@ -48,13 +48,18 @@ def __init__(
...
+            self.neck = DETECTION_NECKS.get(neck_name)(in_channels=neck_in_channels, **neck_params)
+            head_in_channels = self.neck.out_channels
+        else:
+            self.neck = nn.Identity()
+            head_in_channels = self.backbone.out_encoder_channels[-self.num_scales:][::-1]
 
         # HEAD
         head_params = head_params or dict()
-        self.bbox_head = HEADS.get(head_name)(in_channels=self.neck.out_channels, **head_params)
+        self.bbox_head = HEADS.get(head_name)(joint_loss=self.losses, in_channels=head_in_channels, **head_params)
```
PososikTeam, line 62: So we send joint loss to bbox_head, ok

torchok/tasks/detection.py
```diff
@@ -129,21 +141,26 @@ def validation_step(self, batch: Dict[str, torch.Tensor], batch_idx: int,
...
 
         output['prediction'] = self.bbox_head.get_bboxes(**output)
         output['target'] = [dict(bboxes=bb, labels=la) for bb, la in zip(output['gt_bboxes'], output['gt_labels'])]
-        self.metrics_manager.update(Phase.VALID, **output)
-        output_dict = {'loss': total_loss}
-        output_dict.update(tagged_loss_values)
+        self.metrics_manager.update(Phase.VALID, dataloader_idx, **output)
+
+        if self._hparams.task.compute_loss_on_valid:
+            total_loss, tagged_loss_values = self.bbox_head.loss(self.losses, **output)
```
PososikTeam, line 150: But now when we compute the loss we also send joint loss, why?

torchok/models/heads/detection/fcos.py
Animatory: For compatibility

torchok/models/heads/detection/detr.py
```diff
@@ -0,0 +1,377 @@
...
+                img_meta = img_metas[img_id]
+                img_h, img_w = self.get_fit_image_size(img_meta['img_shape'], img_meta['orig_img_shape'])
+                dh = input_img_h - img_h
+                dw = input_img_w - img_w
+
+                masks[img_id, dh // 2:input_img_h + (-dh) // 2, dw // 2: input_img_w + (-dw) // 2] = 0
+            # interpolate masks to have the same spatial shape with x
+            masks = F.interpolate(masks.unsqueeze(1), size=x.shape[-2:]).to(torch.bool).squeeze(1)
+        else:
+            masks = torch.zeros((batch_size, *x.shape[-2:]), dtype=torch.bool, device=x.device)
```
PososikTeam, line 177: Why it should be zeros? Don't see this code in mmdetection

torchok/models/heads/detection/detr.py
```diff
@@ -0,0 +1,377 @@
...
+            factors = torch.cat(factors, 0)
+
+            # DETR regress the relative position of boxes (cxcywh) in the image,
+            # thus the learning target is normalized by the image size. So here
+            # we need to re-scale them for calculating IoU loss
+            bbox_preds = bbox_preds.reshape(-1, 4)
+            bboxes = bbox_cxcywh_to_xyxy(bbox_preds) * factors
+            bboxes_gt = bbox_cxcywh_to_xyxy(bbox_targets) * factors
+
+            total_loss, tagged_loss_values = joint_loss(
```
PososikTeam, line 278: Did you try to return list of all staff that need joint loss to task and compute loss in task? Or it's bad idea because we return list of tensors and it can increase total memory? Or explain reason that it is bad to do it.

torchok/models/heads/detection/detr.py
Animatory: This mask contains an inverted area of interest. If you provide image meta, we can obtain padded image borders and consider them as an area out of our interest, marking them as True or 1. Otherwise, if we don't get meta, we have to consider the whole image as an area of interest marking them as False or 0

torchok/tasks/detection.py
PososikTeam: Add comment please that some detection head can use joint loss parameters and change it

PososikTeam (approved):

PR #186, opened 2023-02-09
Update pytorch-lightning to 1.9.0

Animatory:


torchok/__main__.py
```diff
@@ -31,6 +32,8 @@ def entrypoint(config: DictConfig):
     # Seed everything
     if config.seed_params is not None:
         seed_everything(**config.seed_params)
+    # Use float32 datatype for the internal precision of float32 matrix multiplications.
+    torch.set_float32_matmul_precision("highest")
```
PososikTeam, line 36: I think `highest` is default. Why you add this? If you want change you should add it to config.

PososikTeam: Will add to new version

PososikTeam (approved):

PR #189, opened 2023-02-24
Add weighted samplers

kristinarakova:
Added [WeightedRandomSampler](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/sampler.py) from PyTorch


torchok/constructor/constructor.py
```diff
@@ -281,22 +281,42 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
...
             return dataloaders
         else:
             return []
 
     @staticmethod
-    def _prepare_dataloader(dataset_params: DictConfig, dataloader_params: DictConfig) -> DataLoader:
+    def _prepare_sampler(dataset: ImageDataset, sampler_params: DictConfig) -> Sampler:
+
+        if sampler_params:
+            csv_column = sampler_params.csv_column if sampler_params.get('csv_column') else None
```
PososikTeam, line 301: It's not appropriate way to use some internal parameters in constructor.

torchok/data/datasets/classification/classification.py
```diff
@@ -91,6 +91,7 @@ def __init__(self,
         self.multilabel = multilabel
         self.lazy_init = lazy_init
         self.csv_path = csv_path
+        self.weight_column = 'weight'
```
PososikTeam, line 94: Hard code

torchok/data/datasets/classification/classification.py
```diff
@@ -122,6 +123,11 @@ def get_raw(self, idx: int) -> dict:
 
         return sample
 
+    def get_sampler_weights(self, weight_column=None):
```
PososikTeam, line 126: I think this method should be in Sampler

PososikTeam (changes requested):
- Let's create our own `WeightedSampler` which will be inherit from pytorch `WeightedRandomSampler` and one of its parameter would be `csv_path` with weights.
- Also you need to write structure of `sampler_params` in `config_structure.py` file. I believe that `sampler_params` can be a part of `dataset_params`

torchok/constructor/constructor.py
```diff
@@ -291,12 +291,27 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
             return []
 
     @staticmethod
-    def _prepare_dataloader(dataset_params: DictConfig, dataloader_params: DictConfig) -> DataLoader:
+    def _prepare_sampler(dataset: ImageDataset, sampler_params: DictConfig) -> Sampler:
```
PososikTeam, line 294: This method does't need. In our philosophy if you have parameters in config for some class you can not add additional parameters inside the code. So the sampler must be create only with options like:
```
sampler = SAMPLERS.get(sampler_params.name)(**sampler_params.params)
```
This philosophy makes torchok more simplify to understand, and code become more readable.
Also not each Sampler has parameter for weights. So we wouldn't have ability to add new Samplers.

For create sampler like in code above, you need to create own Sampler Weighted class which inherit from torch Sampler. And it's parameters should be weights_csv_path and weights_column for take the weights.

torchok/data/datasets/classification/classification.py
```diff
@@ -40,6 +40,7 @@ def __init__(self,
                  input_dtype: str = 'float32',
                  target_column: str = 'label',
                  target_dtype: str = 'long',
+                 weight_column: str = 'weight',
```
PososikTeam, line 43: Sampler is different entity, we need split dataset entity and sampler entity. This parameter can't be here.

torchok/data/datasets/classification/classification.py
```diff
@@ -122,6 +125,11 @@ def get_raw(self, idx: int) -> dict:
 
         return sample
 
+    def get_sampler_weights(self):
```
PososikTeam, line 128: Sampler is different entity, we need split dataset entity and sampler entity. This function can't be here. Move it to Sampler class.

torchok/constructor/config_structure.py
```diff
@@ -57,12 +57,19 @@ class AugmentationParams:
...
+    params: Dict = field(default_factory=dict)
+
+
 @dataclass
 class DatasetParams:
     name: str
     params: Dict
     transform: List[AugmentationParams]
     augment: Optional[List[AugmentationParams]] = field(default_factory=list)
+    sampler: Optional[SamplerParams] = None
```
PososikTeam, line 72: Sampler must be Dataloader parameter not Dataset

torchok/constructor/constructor.py
```diff
@@ -281,22 +281,31 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
...
         """
         if phase in self.hparams.data:
-            dataloaders = [
-                self._prepare_dataloader(phase_params.dataset, phase_params.dataloader)
-                for phase_params in self.hparams.data[phase] if phase_params is not None
-            ]
+            dataloaders = []
+            for phase_params in self.hparams.data[phase]:
+                if phase_params is not None:
+                    sampler_params = phase_params.sampler if phase_params.get('sampler') else None
```
PososikTeam, line 287: you can write `phase_params.get('sampler')` if 'sampler not in dict it return None

torchok/constructor/constructor.py
```diff
@@ -281,22 +281,31 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
             - ValueError: When OneOrOther composition is passed that isn't supported
         """
         if phase in self.hparams.data:
-            dataloaders = [
-                self._prepare_dataloader(phase_params.dataset, phase_params.dataloader)
```
PososikTeam, line 285: just add `phase_params.get('sampler')` to parameter

torchok/data/samplers/WeightedSampler.py
```diff
@@ -0,0 +1,35 @@
...
+
+        Args:
+            data_folder: Directory with csv file.
+            csv_path: Path to the csv file with path to images and annotations.
+            weight_column: column name containing weights for each image.
+    """
+    def __init__(self,
+                 data_folder: str,
+                 csv_path: str,
+                 weight_column: str,
```
PososikTeam, line 19: write default value for `weight_column` with the name which you use

torchok/constructor/constructor.py
```diff
@@ -281,22 +281,33 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
...
+                            dataloader_params: DictConfig,
+                            sampler_params: DictConfig) -> DataLoader:
         dataset = Constructor._create_dataset(dataset_params)
         collate_fn = dataset.collate_fn if hasattr(dataset, 'collate_fn') else None
-
+        if sampler_params is not None:
+            num_samples = sampler_params.params.get('num_samples', len(dataset))
+        sampler = (
+            SAMPLERS.get(sampler_params.name)(num_samples=num_samples, **sampler_params.params)
+            if sampler_params is not None else None
```
PososikTeam, line 306: By default it's dict it can't be None, you can skip this check

```
@dataclass
class SamplerParams:
    name: str
    params: Dict = field(default_factory=dict)
```

torchok/constructor/constructor.py
```diff
@@ -281,22 +281,33 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
...
     @staticmethod
-    def _prepare_dataloader(dataset_params: DictConfig, dataloader_params: DictConfig) -> DataLoader:
+    def _prepare_dataloader(dataset_params: DictConfig,
+                            dataloader_params: DictConfig,
+                            sampler_params: DictConfig) -> DataLoader:
         dataset = Constructor._create_dataset(dataset_params)
         collate_fn = dataset.collate_fn if hasattr(dataset, 'collate_fn') else None
-
+        if sampler_params is not None:
+            num_samples = sampler_params.params.get('num_samples', len(dataset))
```
PososikTeam, line 303: you need pop this because, if `sampler_params.params` has `num_samples` you put twice num_samples parameter to constructor
```
SAMPLERS.get(sampler_params.name)(num_samples=num_samples, **sampler_params.params)
```
But I use code for situations like this
```
if 'num_samples' not in sampler_params:
    sampler_params.params['num_samples'] = len(dataset)
```
In this case you not remove useful data from DictConfig. Suggest you to use this code.

torchok/constructor/constructor.py
```diff
@@ -281,22 +281,33 @@ def create_dataloaders(self, phase: Phase) -> List[DataLoader]:
             - ValueError: When OneOrOther composition is passed that isn't supported
         """
         if phase in self.hparams.data:
-            dataloaders = [
```
PososikTeam, line 284: Please return Vlads code =)
```
dataloaders = [
                self._prepare_dataloader(phase_params.dataset, phase_params.dataloader, phase_params.get('sampler')
                for phase_params in self.hparams.data[phase] if phase_params is not None
            ]
 ```
 It has 4 lines and more readable i think.

PososikTeam (approved):

PR #198, opened 2023-05-05
Update pytorch-2.0 compatibility

Animatory:


torchok/constructor/constructor.py
```diff
@@ -329,10 +329,8 @@ def _prepare_transforms_recursively(transforms: ListConfig) -> List[Union[A.Comp
             transform_name = transform_info.name
             transform_params = transform_info.get('params', dict())
 
-            if transform_name in ['Compose', 'OneOf', 'SomeOf', 'PerChannel', 'Sequential']:
+            if 'transforms' in transform_params:
```
VladVin, line 332: It's hard to understand the logic behind this. Why did you change the original implementation?

torchok/models/heads/detection/__init__.py
```diff
@@ -1,6 +1,7 @@
 import importlib
 
 has_mmcv = importlib.util.find_spec("mmcv")
-if has_mmcv is not None:
-    import torchok.models.heads.detection.fcos
-    import torchok.models.heads.detection.detr
+# ToDo: fix mmdet heads
```
VladVin, line 4: So, detection doesn't work?

torchok/tasks/base.py
```diff
@@ -121,17 +122,18 @@ def on_predict_start(self) -> None:
         if self._hparams.task.load_checkpoint is not None:
             load_checkpoint(self, **self._hparams.task.load_checkpoint)
 
-    def training_step(self, batch: Dict[str, Union[torch.Tensor, int]], batch_idx: int) -> Dict[str, torch.Tensor]:
+    def training_step(self, batch: Dict[str, Union[Tensor, int]], batch_idx: int) -> Dict[str, Tensor]:
         """Complete training loop."""
         output = self.forward_with_gt(batch)
         total_loss, tagged_loss_values = self.losses(**output)
+        self.log("loss", total_loss, prog_bar=True, on_step=True)
```
VladVin, line 129: We don't usually log on each step. Why did you decide it to do for user?

VladVin (changes requested):

torchok/constructor/config_structure.py
```diff
@@ -134,17 +134,12 @@ class TaskParams:
 # Trainer parameters
 @dataclass
 class TrainerParams:
-    enable_checkpointing: bool = True
-    default_root_dir: Optional[str] = None
-    gradient_clip_val: Optional[float] = None
-    gradient_clip_algorithm: Optional[str] = None
+    accelerator: str = "auto"
+    strategy: str = "auto"
+    devices: str = "auto"
```
VladVin, line 139: Shouldn't be the example configs updated after all these changes?

VladVin (changes requested):

.github/workflows/cd.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.8", "3.9", "3.10"]
```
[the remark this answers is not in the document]
Animatory: It is time to move forward, and PyTorch 2.0 does not support 3.7

torchok/constructor/constructor.py
Animatory: Because before, we were limited to adding new container augmentations. The common thing for all container transforms is the 'transforms' parameter requirement. So, now we can add custom container augmentations

torchok/models/heads/detection/__init__.py
Animatory: Yes, but I am fixing this

torchok/constructor/config_structure.py
Animatory: No, they are compatible

torchok/constructor/constructor.py
VladVin: Please add a comment about it here

torchok/models/heads/detection/__init__.py
VladVin: What does it require to make it work? I'm estimating whether to put it into this PR

torchok/tasks/base.py
VladVin: But we didn't track loss on each single step. It should be tracked with the configured frequency. So, how do configure that?

Animatory: Your CMD logs included step loss. The step size is determined in the config in TQDMProgressBar callback. Loss was tracked automatically in the previous version, but not it is not.

torchok/constructor/constructor.py
Animatory: Added

torchok/tasks/base.py
Animatory: Because I want loss to be displayed at every step

VladVin (approved):

